PR #11, opened 2014-08-03
Remove the default table prefix (Fix #5)

### LINKED ISSUE #5: Remove 'jos_' default prefix
We need to allow for no database prefix here:

https://github.com/joomla-framework/database/blob/master/src/DatabaseDriver.php#L401

This causes a b/c problem so we should sync this with a 2.0 release.

mbabker:
See #5

### DIFF
--- a/Tests/DatabaseMysqlCase.php
+++ b/Tests/DatabaseMysqlCase.php
@@ -20,7 +20,7 @@ abstract class DatabaseMysqlCase extends TestDatabase
 	 * @var    array  The database driver options for the connection.
 	 * @since  1.0
 	 */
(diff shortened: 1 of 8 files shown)


mbabker:
@wilsonge @eddieajau Yay or nay?  This would be rebased to the 2.0 branch if yay.

wilsonge:
yay

eddieajau:
Can you add a test to check that an empty prefix is allowed?

mbabker:
Done

eddieajau:
If it passes, it's a :+1: from me.

PR #12, opened 2014-08-17
Fix union() query

asika32764:
Since we remove union clause  from select, we have to add union to `type` that it can work.

And a union type should not begin with `UNION (`


photodude:
If this is a necessary change, you need to fix the unit tests as well.

csthomas:
Take a look  and test a new PR https://github.com/joomla-framework/database/pull/116

asika32764:
Close since this PR is out-of-dated.

PR #15, opened 2015-03-22
Package Review

mbabker:
Pertinent changes:

- Removes overridden getQuery method in PostgreSQL driver
- Removes deprecated method in PostgresqlImporter
- DatabaseQuery will trigger an error when accessing an undefined property via magic getter
- Uses the DatabaseFactory within the DatabaseDriver
- Organizes the tests folder


wilsonge:
Looks good at a code review. Will wait for travis before merging.

wilsonge:
Is it worth adding a method to set the database type in the factory and then making the name params optional? So you can do

```
$factory->setDatabaseType('mysqli');
// Automatically return Joomla\Database\Mysqli\MysqliDriver
$factory->getDriver();
// Automatically return Joomla\Database\Mysqli\MysqliQuery
$factory->getQuery();
```
etc

mbabker:
PR it separately.  I can't think of a reason not to.

PR #17, opened 2015-05-13
Run tests against php 7 and HHVM

photodude:
- Allow php 7 and HHVM to fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,13 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
+  - hhvm
+
+matrix:
+  allow_failures:
+    - php: hhvm
+    - php: 7.0
 
 before_script:
   - composer update


mbabker:
We're actually doing this on all the package 2.0 branches already (https://github.com/joomla-framework/database/blob/2.0-dev/.travis.yml).

photodude:
@mbabker Is there any reason not to do this on the existing branches?

mbabker:
I was thinking at the time to focus on 2.0 dev and B/C breaks could be more easily dealt with there compared to 1.x.  Not to say 1.x code doesn't work with HHVM or PHP7 but it's easier to deal with compatibility issues in 2.0 then back port as able (for me at least).

photodude:
I can see that. 
I was thinking since I think the CMS uses the 1.x branches of the framework currently, making the php7 and hhvm tests visible in the 1.x branches might be helpful. It would also help when/if B/C back ports from the 2.x branch occurs.

mbabker:
Ya good point

photodude:
Closing as this was implemented via a different PR

PR #22, opened 2015-07-07
Use the PSR logging for missing method

wilsonge:


src/DatabaseImporter.php
```diff
@@ -272,10 +272,11 @@ public function mergeStructure()
 						}
 						catch (\RuntimeException $e)
 						{
-							$this->addLog('Fail: ' . $this->db->getQuery());
+							$this->db->log(LogLevel::DEBUG, 'Fail: ' . $this->db->getQuery());
```
mbabker: Missing the use statement for this class

wilsonge: Oops

PR #23, opened 2015-07-07
Add method as abstract

wilsonge:
This method is currently implemented by all subclasses as it's called in the ```mergeStructure``` method. This just formalises it

### DIFF
--- a/src/DatabaseImporter.php
+++ b/src/DatabaseImporter.php
@@ -139,6 +139,17 @@ abstract class DatabaseImporter
 	}
 
 	/**
+	 * Get alters for table if there is a difference.
+	 *
+	 * @param   \SimpleXMLElement  $structure  The XML structure pf the table.
+	 *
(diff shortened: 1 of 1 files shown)


mbabker:
It's borderline questionable as it technically could breach B/C (highly hypothetical but a custom driver that subclasses the importer and overrides mergeStructure to not call that method).

wilsonge:
Up to you to decide on. If you wanna leave it for 2.x go for it

mbabker:
So here's the thing.  Technically adding a new abstract method to a base abstract class or a new method to an interface would cause a B/C breach as it adds a new implementation requirement to subclasses.  Granted it's basically a requirement now in the base class, but one that can be bypassed in a subclass.  Adding the abstract method can't be bypassed.

By that argument, a 1.x "fix" would be checking if method exists and throwing an exception and the abstract declaration shifted for 2.0.

PR #24, opened 2015-07-08
Add missing xmlToCreate function

wilsonge:
Used at https://github.com/joomla-framework/database/blob/master/src/DatabaseImporter.php#L288

### DIFF
--- a/src/Mysqli/MysqliImporter.php
+++ b/src/Mysqli/MysqliImporter.php
@@ -49,8 +49,8 @@ class MysqliImporter extends DatabaseImporter
 	 *
 	 * @return  string
 	 *
-	 * @since   3.4.4
-	 * @throws  RuntimeException
+	 * @since   __DELPOY_VERSION__
+	 * @throws  \RuntimeException
 	 */
 	protected function xmlToCreate(SimpleXMLElement $table)
 	{
@@ -59,7 +59,7 @@ class MysqliImporter extends DatabaseImporter
 
 		if (in_array($tableName, $existingTables))
 		{
-			throw new RuntimeException('The table you are trying to create already exists');
+			throw new \RuntimeException('The table you are trying to create already exists');
 		}
 
 		$createTableStatement = 'CREATE TABLE ' . $this->db->quoteName($tableName) . ' (';
--- a/src/Mysqli/MysqliImporter.php
+++ b/src/Mysqli/MysqliImporter.php
@@ -76,7 +76,7 @@ class MysqliImporter extends DatabaseImporter
 
 		$createTableStatement = rtrim($createTableStatement, ', ');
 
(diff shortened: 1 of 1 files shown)


src/Mysqli/MysqliImporter.php
```diff
@@ -43,6 +43,45 @@ public function check()
...
+		}
+
+		foreach ($table->xpath('key') as $key)
+		{
+			$createTableStatement .= $this->getKeySQL(array($key)) . ', ';
+		}
+
+		$createTableStatement = rtrim($createTableStatement, ', ');
+
+		$createTableStatement .= ') ENGINE=InnoDB DEFAULT CHARSET=utf8';
```
mbabker: We shouldn't be hardcoding engines in here.

wilsonge: Where should be allow that to be set. There is nothing in the exporter that grabs the database information presently (which if we were to add should go into another PR). It's probably the same for the charset tbh. we have a method to see if it's supported in the database engine level - but nothing for the table level....

mbabker: So aim for a larger PR that deals with all the importer/exporter shortcomings instead of trying to piecemeal it all together in bits and pieces; it's already broken anyway so let's just focus on getting it working altogether.  As I suggested in the CMS pull just leave the engine out for the moment until you have a method to get that extended data.

wilsonge: Surely that would have to be a framework v2 thing though? Does this not work well enough for Framework v1?

mbabker: With the hardcoded stuff in it, I say no.  Refer to the last sentence on my last comment :wink:

wilsonge: I've removed the engine specific stuff. Do we need to remove the default charset as well. I guess this function suggests that technically we don't require it https://github.com/joomla-framework/database/blob/master/src/DatabaseDriver.php#L779-L789

mbabker: Is our code usable with other charsets?  If so, I'd drop it.  If our code only works with UTF-8, well, I'd say that might be a bug to address separately.

wilsonge: I have no clue. I mean the project I'm using isn't storing UTF-8 chars all the time for the chinese in some rows of the table - so I guess it works (although maybe that's why some pages have a load time of 20 years)

mbabker: Just leave it.  Before I drop my head any harder onto my desk over the WTF moments I'm having.

wilsonge: lmao. I remove the default charset. Can we just merge this into v1 and then do the full rewrite?

PR #26, opened 2015-07-08
Remove prefix replacing methods in favour of DatabaseDriver::replacePrefix

wilsonge:


wilsonge:
So the tests are actually showing that the prefix isn't currently replaced (which actually I can understand in the output). Maybe we just should be calling this method at all. It clearly doesn't actually replace the prefix.

mbabker:
I'd say the tests are bad then.  Presumably if you're exporting the table structure, you'd want the full table name and not our magic prefix at the beginning (since at least for the MySQL exporter it's dumping in their standard XML format).  If we were exporting to a "Joomla" format, you could make an argument for the prefix.

wilsonge:
So it's actually more complex than I first realised. The exporter converts a prefixed name (e.g. ```jos_test``` into a generic prefixed name e.g. ```#__test```. So currently it's 100% intended that this is the case. It's not an accident or bad code. It's the intended result

The importer method actually does the same thing as ```DatabaseDriver::replacePrefix()```

mbabker:
I quit :facepalm: :headbang: :wtf:

PR #27, opened 2015-07-08
Exceptions cannot be thrown in magic __toString

wilsonge:
This is a PHP requirement. If an exception goes uncaught in a ```__toString()``` magic method then a PHP Fatal Error is generated

### DIFF
--- a/src/DatabaseExporter.php
+++ b/src/DatabaseExporter.php
@@ -104,7 +104,7 @@ abstract class DatabaseExporter
 					break;
 			}
 		}
-		catch (Exception $e)
+		catch (\Exception $e)
 		{
 			// Do nothing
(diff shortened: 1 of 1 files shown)


src/DatabaseExporter.php
```diff
@@ -85,20 +85,28 @@ public function __construct()
...
+			// Get the format.
+			switch ($this->asFormat)
+			{
+				case 'xml':
+				default:
+					$buffer = $this->buildXml();
+					break;
+			}
+		}
+		catch (Exception $e)
```
mbabker: Need the leading slash on Exception

wilsonge: Fixed

PR #31, opened 2015-09-27
Improve availability of error data on connect errors

mbabker:
This PR improves the error handling in the driver `connect()` methods by adding logging for verbose error data when available.

### DIFF
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -177,7 +177,7 @@ class MysqliDriver extends DatabaseDriver
 		// Attempt to connect to the server.
 		if (!$this->connection)
 		{
-			$this->log('error', 'Could not connect to MySQL: ' . mysqli_connect_error());
+			$this->log(Log\LogLevel::ERROR, 'Could not connect to MySQL: ' . mysqli_connect_error());
 
 			throw new \RuntimeException('Could not connect to MySQL.', mysqli_connect_errno());
 		}
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -177,7 +177,9 @@ class MysqliDriver extends DatabaseDriver
 		// Attempt to connect to the server.
 		if (!$this->connection)
 		{
-			throw new \RuntimeException('Could not connect to MySQL.');
+			$this->log('error', 'Could not connect to MySQL: ' . mysqli_connect_error());
+
+			throw new \RuntimeException('Could not connect to MySQL.', mysqli_connect_errno());
 		}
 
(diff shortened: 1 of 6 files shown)


src/Mysqli/MysqliDriver.php
```diff
@@ -177,7 +177,9 @@ public function connect()
 		// Attempt to connect to the server.
 		if (!$this->connection)
 		{
-			throw new \RuntimeException('Could not connect to MySQL.');
+			$this->log('error', 'Could not connect to MySQL: ' . mysqli_connect_error());
+
+			throw new \RuntimeException('Could not connect to MySQL.', mysqli_connect_errno());
```
wilsonge: Do we not want to add the mysql connect error in here as well?

src/Oracle/OracleDriver.php
```diff
@@ -94,11 +94,6 @@ public function __destruct()
 	 */
 	public function connect()
 	{
-		if ($this->connection)
```
wilsonge: I'm might be having a mind blank here - but I thought if this check returns in the parent method then won't the rest of this method will still run? It just aborts the parent method

mbabker: If it does then PHP is FUBAR.  But the parent calling return should return too.  Otherwise this needs to be fixed in the PDO MySQL class.

src/Mysqli/MysqliDriver.php
mbabker: I didn't change the message text explicitly since the exception may be getting displayed directly in applications (I know in the CMS this exception does end up being rendered by the error handler).  I guess it depends on what text is given by the `mysqli_connect_error()` function but this erred toward a more cautious approach.

src/Oracle/OracleDriver.php
wilsonge: Well I'm just thinking normally when you call a parent function you normally do

`return parent::load` or whatever rather than just `parent::load`

mbabker: You only call it as `return parent::whatever` when you intend to return the value of the parent method call (it's the same as when you call `return $this->whatever` within a method).

wilsonge: http://php.net/manual/en/keyword.parent.php#42153

wilsonge: What i'm trying to say is that sure `PDODriver::__construct()` will return at the point you check for the connection. But the rest of `OracleDriver::__construct()` will run. I think.

mbabker: Now you're getting methods crossed.  This isn't the constructor :wink: 

And PHP is stupid

PR #38, opened 2016-06-29
Prepared Statement Support for MySQLi Driver

mbabker:
Exposes support for prepared statements to the MySQLi driver by way of the existing `Query\PreparableInterface` API.

Note this API was originally created for PDO based drivers so it has a "funny" API for our non-PDO code.  None the less it's still usable.

Also consistently uses the OOP style `mysqli` API versus procedural since we're storing an object anyway.  Only in methods where a cursor can be injected does the procedural style remain.

Support for PostgreSQL and SQL Server will be forthcoming.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,10 +38,8 @@ matrix:
 before_script:
   - composer self-update
   - composer update $COMPOSER_FLAGS
-  - if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then mysql -e 'create database joomla_ut;'; fi
-  - if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then mysql joomla_ut < Tests/Stubs/mysql.sql; fi
-  - if [[ $TRAVIS_PHP_VERSION = hhv* ]]; then mysql -u root -e 'create database joomla_ut;'; fi
-  - if [[ $TRAVIS_PHP_VERSION = hhv* ]]; then mysql -u root joomla_ut < Tests/Stubs/mysql.sql; fi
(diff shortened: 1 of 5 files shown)


franzpeter:
Hello Michael, in case of prepared statements for MySQLi what will it return? Single fetch step by step or, sorry for my bad english, something like a list of all fetched? The single method is working step by step without producing memory issues? Will that be possible?

mbabker:
It depends on how you execute the query.  The returns should be backward compatible to the existing implementation so there should be zero change from a public API's use case (so something like `loadObjectList()` should still load all results as happens now and an iterator should still let you load each result set step by step).

franzpeter:
Hello Michael: Simple question about mysqli prepared statements. It is necessary for mysqli to provide the params type like: s, i, d or b per column. Would it be possible to use getTableColumns to fill the type automatically? I mean it should translate by the given result of getTableColumns. Example: table has column types defined as date, varchar, char, text. So all are of type s. If there are any kind of int it is i, for decimals it is d. And because getTableColumns contains the names too it could create those variables too by taking the name and make it as $name. Maybe I am talking nonsense but just an idea.

franzpeter:
I think that MySQLWorkbench does somehow the same if you enter a query and use Copy as PHP code (iterate) from the tools tab of that app.

mbabker:
Right now you should specify types.  To be honest, the API for prepared statements is something I want to look at for the 2.0 branch since the existing API was written specifically for the PDO drivers and that doesn't translate over to MySQLi or "plain" PostgreSQL as well (plus there's some portability issues since right now the query syntax relies on the native driver implementation).

PR #47, opened 2016-08-24
Fix lockTable and unlockTables methods

jbanety:


mbabker:
Can you submit this on master?

jbanety:
Aie!

PR #56, opened 2016-09-17
Disable xdebug since travis is not doing code coverage.

photodude:
Merge by code review, composer should no longer bark about xdebug running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,12 @@ matrix:
     - php: hhvm
 
 before_script:
+  # Disable xdebug on php 7.0.* and lower.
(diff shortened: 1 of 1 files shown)


mbabker:
So two things:

1) How much are we really gaining here turning off XDebug?
2) At some point I'd like to have all of the repos wired in with Scrutinizer (some of them are), and that does collect code coverage.

photodude:
1) This cuts a few seconds off the travis builds, possibly more.
2) This only affects the travis builds
3) +1 for Scrutinizer,

mbabker:
The way we have things set up (see https://github.com/joomla-framework/date/blob/2.0-dev/.travis.yml as an example), the coverage is collected on Travis and sent to Scrutinizer.  So we'd really just be reverting this as soon as we finished setting things up (if it happens).

photodude:
So we'll need to re-enable xdebug on the CMS if/when Scrutinizer is set up there....

Issue #54, opened 2016-09-17
Postgresql Driver Execution errors on insert


photodude:
The Postgresql Driver has Execution errors on insert with the following the unit tests on travis

``` php
1) Joomla\Database\Tests\DriverPostgresqlTest::testExecute
Joomla\Database\Exception\ExecutionFailureException: 
SQL=INSERT INTO jos_dbtest
(title,start_date,description) VALUES 
('testTitle','1970-01-01','testDescription')
/database/src/Postgresql/PostgresqlDriver.php:819
/database/Tests/DriverPostgresqlTest.php:1131

2) Joomla\Database\Tests\DriverPostgresqlTest::testExecutePreparedStatement
Joomla\Database\Exception\ExecutionFailureException: 
SQL=INSERT INTO jos_dbtest
(title,start_date,description) VALUES 
($1, $2, $3)
/database/src/Postgresql/PostgresqlDriver.php:819
/database/Tests/DriverPostgresqlTest.php:1160
```

photodude:
There are also the following IteratorPgsqlTest failures not listed in travis due to JTEST_DATABASE_PGSQL_DSN having been commented out in phpunit.travis.xml
```php
There were 5 failures:
1) Joomla\Database\Tests\IteratorPgsqlTest::testForEach with data set #0 ('title', '#__dbtest', null, 'stdClass', 0, 0, array(stdClass Object (...), stdClass Object (...), stdClass Object (...), stdClass Object (...)), null)
136
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => stdClass Object (
-        'title' => 'Testing'
+        'title' => 'testTitle'
     )
     1 => stdClass Object (
-        'title' => 'Testing2'
+        'title' => 'Testing'
     )
     2 => stdClass Object (
-        'title' => 'Testing3'
+        'title' => 'Testing2'
     )
     3 => stdClass Object (
-        'title' => 'Testing4'
+        'title' => 'Testing3'
     )
+    4 => stdClass Object (...)
 )
/database/Tests/IteratorPgsqlTest.php:137

2) Joomla\Database\Tests\IteratorPgsqlTest::testForEach with data set #1 ('title', '#__dbtest', null, 'stdClass', 2, 0, array(stdClass Object (...), stdClass Object (...)), null)
136
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => stdClass Object (
[...]

Issue #53, opened 2016-09-17
Tests causing build errors due to testing with JTEST_DATABASE_PGSQL_DSN


photodude:
The following tests in JTEST_DATABASE_PGSQL_DSN are causing php unit to fail to complete testing;
due to these tests JTEST_DATABASE_PGSQL_DSN has been commented out in phpunit.travis.xml.
- DriverPostgresqlTest::testExecutePreparedStatement
- DriverPostgresqlTest::testRenameTable
- DriverPostgresqlTest::testTransactionStart
- DriverPostgresqlTest::testTransactionCommit
- DriverPostgresqlTest::testInsertObject
- DriverPostgresqlTest::testInsertid
- DriverPgsqlTest::testRenameTable

These tests could be excluded from the test suit until they are fixed so more tests are completed.

For example, once the tests are excluded the following IteratorPgsqlTest failures are reported

``` php
There were 5 failures:
1) Joomla\Database\Tests\IteratorPgsqlTest::testForEach with data set #0 ('title', '#__dbtest', null, 'stdClass', 0, 0, array(stdClass Object (...), stdClass Object (...), stdClass Object (...), stdClass Object (...)), null)
136
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => stdClass Object (
-        'title' => 'Testing'
+        'title' => 'testTitle'
     )
     1 => stdClass Object (
-        'title' => 'Testing2'
+        'title' => 'Testing'
     )
     2 => stdClass Object (
-        'title' => 'Testing3'
+        'title' => 'Testing2'
     )
     3 => stdClass Object (
-        'title' => 'Testing4'
+        'title' => 'Testing3'
     )
+    4 => stdClass Object (...)
 )
/database/Tests/IteratorPgsqlTest.php:137

2) Joomla\Database\Tests\IteratorPgsqlTest::testForEach with data set #1 ('title', '#__dbtest', null, 'stdClass', 2, 0, array(stdClass Object (...), stdClass Object (...)), null)
136
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => stdClass Object (
-        'title' => 'Testing'
+        'title' => 'testTitle'
     )
     1 => stdClass Object (
-        'title' => 'Testing2'
+        'title' => 'Testing'
     )
 )
/database/Tests/IteratorPgsqlTest.php:137

[...]

PR #60, opened 2016-10-23
Fix the PDO cursors

photodude:
Pull Request for Issues #53 and #54

### Summary of Changes
$driverOptions should default to array(\PDO::ATTR_CURSOR => \PDO::CURSOR_SCROLL)
http://www.dragonbe.com/2015/07/speeding-up-database-calls-with-pdo-and.html

> we need to ensure the cursor is now scrollable so we can use row by row control.

### Testing Instructions
Review Travis CI tests, we now have successful tests

Any real world use of the database package should work after applying this patch

We many need to specify that the $driverOptions must include `\PDO::ATTR_CURSOR => \PDO::CURSOR_SCROLL` in the `array()` to ensure the cursor is now scrollable so we can use row by row control

### DIFF
--- a/phpunit.travis.xml
+++ b/phpunit.travis.xml
@@ -4,7 +4,7 @@
 	<php>
 		<const name="JTEST_DATABASE_MYSQL_DSN" value="host=localhost;dbname=joomla_ut;user=root;pass=" />
 		<const name="JTEST_DATABASE_MYSQLI_DSN" value="host=localhost;dbname=joomla_ut;user=root;pass=" />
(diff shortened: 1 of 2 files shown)


photodude:
I may have been too optimistic... Just noticed all tests (although green) are actually failing due to `Recursion trying to check if connected`

photodude:
Never mind.... PR #60 is a failure After skipping some of the tests that reported `Recursion trying to check if connected` I still got the existing errors.

PR #62, opened 2016-12-03
Fix Postgresql DriverTest errors by adding tearDown()

photodude:
Pull Request for Issue #54

### Summary of Changes
add teardown to test to ensure tests run isolated

### Testing Instructions
merge by code review, travis tests pass

none

### DIFF
--- a/Tests/DriverPostgresqlTest.php
+++ b/Tests/DriverPostgresqlTest.php
@@ -1235,4 +1235,17 @@ class DriverPostgresqlTest extends DatabasePostgresqlCase
 
 		$this->assertThat($result, $this->equalTo($expected), __LINE__);
 	}
+
+	/**
+	 * Tears down the fixture.
+	 *
+	 * This method is called after a test is executed.
+	 *
+	 * @return void
+	 */
+	protected function tearDown()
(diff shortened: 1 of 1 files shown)


mbabker:
Hmm, maybe we should do this in https://github.com/joomla-framework/test/blob/master/src/TestDatabase.php instead so that all tests are running with an isolated driver.

Either way this'll require other changes because that same TestDatabase class only sets up the driver once in the `setUpBeforeClass()` method then its `setUp()` method will skip tests if the driver isn't there.

photodude:
I see, looks like the skipping is happening with this change. pre-patch there are like 104 skipped, post-patch 165 are skipped.

Making this change in https://github.com/joomla-framework/test/blob/master/src/TestDatabase.php results in a new error
```php
Argument 1 passed to PHPUnit_Extensions_Database_DefaultTester::__construct() must implement interface PHPUnit_Extensions_Database_DB_IDatabaseConnection, null given, called in /vendor/phpunit/dbunit/PHPUnit/Extensions/Database/TestCase.php on line 91 and defined
```

looks like another failed solution to solving these testing issues

PR #69, opened 2017-01-15
Port several changes from the CMS

photodude:
Pull Request for Issue Port several changes from the CMS

### Summary of Changes
- joomla/joomla-cms/#31087 Add SQL queries profiler
- use disconnect method
- https://github.com/joomla/joomla-cms/pull/12359 When $this->cursor is false, throw an Exception
- Disable escape_string_warning option. Fixes https://github.com/joomla/joomla-cms/pull/4157

### Testing Instructions
merge by code review

none

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -204,6 +204,12 @@ abstract class DatabaseDriver implements DatabaseInterface, Log\LoggerAwareInter
 	private $logger;
 
 	/**
+	 * @var    callable[]  List of callables to call just before disconnecting database
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $disconnectHandlers = array();
+
+	/**
 	 * Get a list of available database connectors.  The list will only be populated with connectors that both
 	 * the class exists and the static test method returns true.  This gives us the ability to have a multitude
 	 * of connector classes that are self-aware as to whether or not they are able to be used on a given system.
--- a/src/Postgresql/PostgresqlDriver.php
+++ b/src/Postgresql/PostgresqlDriver.php
@@ -126,10 +126,7 @@ class PostgresqlDriver extends DatabaseDriver
 	 */
 	public function __destruct()
 	{
-		if (is_resource($this->connection))
-		{
-			pg_close($this->connection);
-		}
+		$this->disconnect();
 	}
 
 	/**
@@ -203,6 +200,7 @@ class PostgresqlDriver extends DatabaseDriver
 
 		pg_set_error_verbosity($this->connection, PGSQL_ERRORS_DEFAULT);
 		pg_query($this->connection, 'SET standard_conforming_strings=off');
+		pg_query($this->connection, 'SET escape_string_warning=off');
 	}
 
 	/**
@@ -217,6 +215,11 @@ class PostgresqlDriver extends DatabaseDriver
 		// Close the connection.
 		if (is_resource($this->connection))
 		{
+			foreach ($this->disconnectHandlers as $h)
+			{
+				call_user_func_array($h, array( &$this));
+			}
+
 			pg_close($this->connection);
 		}
 
@@ -330,6 +333,19 @@ class PostgresqlDriver extends DatabaseDriver
 		$array = $this->loadAssocList();
 
 		return $array[0]['lc_collate'];
+	}
+
+	/**
+	 * Method to get the database connection collation, as reported by the driver. If the connector doesn't support
+	 * reporting this value please return an empty string.
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getConnectionCollation()
+	{
+		return pg_client_encoding($this->connection);
 	}
 
 	/**
@@ -723,6 +739,12 @@ class PostgresqlDriver extends DatabaseDriver
 		// Take a local copy so that we don't modify the original query and cause issues later
 		$sql = $this->replacePrefix((string) $this->sql);
 
+
+		if (!($this->sql instanceof LimitableInterface) && ($this->limit > 0 || $this->offset > 0))
+		{
+			$sql .= ' LIMIT ' . $this->limit . ' OFFSET ' . $this->offset;
+		}
+
 		$count = $this->getCount();
 
 		// Increment the query counter.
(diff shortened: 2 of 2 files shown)


src/Postgresql/PostgresqlDriver.php
```diff
@@ -217,6 +215,11 @@ public function disconnect()
 		// Close the connection.
 		if (is_resource($this->connection))
 		{
+			foreach ($this->disconnectHandlers as $h)
```
mbabker: The notion of disconnect handlers does not exist here.

src/Postgresql/PostgresqlDriver.php
```diff
@@ -723,6 +739,12 @@ public function execute()
 		// Take a local copy so that we don't modify the original query and cause issues later
 		$sql = $this->replacePrefix((string) $this->sql);
 
+
+		if (!($this->sql instanceof LimitableInterface) && ($this->limit > 0 || $this->offset > 0))
```
mbabker: Is this actually valid with the rewrites for prepared statements still?

src/Postgresql/PostgresqlDriver.php
```diff
@@ -781,9 +803,15 @@ public function execute()
 				// If connect fails, ignore that exception and throw the normal exception.
 				{
 					// Get the error number and message.
-					$this->errorNum = (int) pg_result_error_field($this->cursor, PGSQL_DIAG_SQLSTATE);
 					$this->errorMsg = pg_last_error($this->connection);
 
+					if ($this->cursor === false)
+					{
+						throw new JDatabaseExceptionExecuting($this->sql, $this->errorMsg);
```
mbabker: Wrong class name

mbabker: Also, this seems borderline incorrect.  What's the difference with the check a few lines above for `(!$this->cursor)` and this check?  The cursor will already be a false object so this effectively prevents getting an error code.

src/DatabaseDriver.php
```diff
@@ -204,6 +204,12 @@
 	private $logger;
 
 	/**
+	 * @var    callable[]  List of callables to call just before disconnecting database
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $disconnectHandlers = array();
```
mbabker: 👎 

If we really intend to introduce this type of feature, this should use the `joomla/event` package and proper event dispatching.  This is honestly a pretty garbage hack accepted by the CMS to avoid using plugins the right way.

Also note that none of the profiling behaviors exist in the Framework code right now either.  There is no use case for this.

src/Postgresql/PostgresqlDriver.php
photodude: It's the change from joomla/joomla-cms#12359, 
```php
pg_result_error_field() expects parameter 1 to be resource, boolean given
```
I previously thought it was ok, but on further consideration of our conversation in joomla/joomla-cms#12359, I believe this is necessary here.

mbabker: I still feel like this entire branch is bugged but whatever.

src/Postgresql/PostgresqlDriver.php
```diff
@@ -744,7 +766,7 @@ public function execute()
 		$this->errorMsg = '';
 
 		// Bind the variables
-		if ($this->sql instanceof PreparableInterface)
+		if ($this->sql instanceof PreparableInterface && !instanceof LimitableInterface)
```
mbabker: The reason for this change?

src/DatabaseDriver.php
photodude: I wasn't exactly sure of why it was added, by @beat in the CMS. I was just trying to bring some consistency between the two libraries. For the moment I'll back out the two related changes with this.

src/Postgresql/PostgresqlDriver.php
mbabker: This would actually break prepared statements in full.  https://github.com/joomla-framework/database/blob/master/src/Postgresql/PostgresqlQuery.php implements both interfaces.

src/DatabaseDriver.php
mbabker: I know it's there.  But the implementation detail IMO is wrong.  If the Framework is going to add something that resembles dispatching events, it should use the event dispatcher proper and not some inlined code to reduce inter-package couplings.  Same goes if someone decides the drivers should also include the inline profiling behaviors, use the profiler package (which I honestly believe shouldn't happen either; hook that to the event system too and dispatch immutable events before and after the query is executed to prevent query manipulation).

mbabker:
One more request.

The log statement happening after https://github.com/joomla-framework/database/pull/69/files#diff-5806d9eac0af9ef4ef47d6725ebc6591R803 needs to be moved up so that the error is always logged, otherwise we have a scenario where an Exception is thrown only if we have a cursor and error code; it's easy enough to set a default 0 code if needed and avoid the extra throw path IMO.

src/Postgresql/PostgresqlDriver.php
photodude: I'm not sure what I was thinking on this change. Probably thought it was needed with the other check for `instanceof LimitableInterface`

src/Postgresql/PostgresqlDriver.php
photodude: fixed

src/Postgresql/PostgresqlDriver.php
photodude: I'm not sure. There is also something similar in https://github.com/joomla-framework/database/blob/master/src/Postgresql/PostgresqlQuery.php#L284-L287

maybe it's not needed

photodude:
Since the default error number is 0 I'll just change the check around

src/Postgresql/PostgresqlDriver.php
mbabker: From what I remember, because of how [PostgresqlDriver::setQuery()](https://github.com/joomla-framework/database/blob/master/src/Postgresql/PostgresqlDriver.php#L925-L948) works we no longer need this check in the `execute` method; `setQuery()` will set the right values into the query object then when cast to string the query object should handle converting that correctly.

So what might be problematic still is the fact we have these `$limit` and `$offset` class vars on the driver object.  Honestly, I would suggest that be deprecated and we standardize setting these values to the query object (I think we're at a point where all of our query objects are implementing `LimitableInterface` and `PreparableInterface` now so it's safe to assume the properties will always be available on the query object).

src/Postgresql/PostgresqlDriver.php
photodude: The whole branch is definitely bugged. Still not sure where the root problem is.

src/Postgresql/PostgresqlDriver.php
photodude: I agree $limit and $offset can probably be moved out of the driver object and into the query object. seems like a more appropriate location.

PR #71, opened 2017-01-17
Fix DriverPostgresqlTest(s)

photodude:
Pull Request for Issue #54
Thanks @csthomas for the fix

### Summary of Changes
Fix testExecute()
Fix testExecutePreparedStatement()

### Testing Instructions
code review 

none

### DIFF
--- a/Tests/DriverPostgresqlTest.php
+++ b/Tests/DriverPostgresqlTest.php
@@ -1122,15 +1122,27 @@ class DriverPostgresqlTest extends DatabasePostgresqlCase
 	 */
 	public function testExecute()
 	{
+		$checkQuery = self::$driver->getQuery(true)
+			 ->select('COUNT(*)')
+			 ->from('#__dbtest');
+		self::$driver->setQuery($checkQuery);
+
+		// Data is from `DatabasePostgresqlCase::getDataSet()`
+		$this->assertThat(self::$driver->loadResult(), $this->equalTo(4), __LINE__);
+
+		// jos_dbtest_id_seq == 2 because of `testInsertObject()`
+		// Reset jos_dbtest_id_seq
+		self::$driver->setQuery('ALTER SEQUENCE jos_dbtest_id_seq RESTART WITH 5')->execute();
+		
 		$query = self::$driver->getQuery(true);
 		$query->insert('jos_dbtest')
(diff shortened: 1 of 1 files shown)


mbabker:
Sooner or later we really need to fix the underlying problem (tests aren't properly setting/resetting global state it seems leaving side effects like this).  But it works, so...

photodude:
I agree there is an underlying test issue. But at least it does not seem to be a driver issue. Still not sure why the table sequence is resetting in the unit test environment even if just the testExecute() was the only test running.

csthomas:
For now I can only say a few words about auto increment. It works like:
- only increment value if you put NULL or skip value for column on `INSERT`
- does not change increment value when you put own value `id =1`,
- does not rollback incement value if you do rollback (there will be a gap)
- does not reset if you delete any or all rows
- it should be reseted on `TRUNCATE table` **but it is not**:
http://stackoverflow.com/questions/13989243/sequence-does-not-reset-after-truncating-the-table

Solution could be to look at `PHPUnit_Extensions_Database_Operation_Factory` class.
And change truncate method:
```php
    public static function CLEAN_INSERT($cascadeTruncates = FALSE)
    {
        return new PHPUnit_Extensions_Database_Operation_Composite(array(
            self::TRUNCATE($cascadeTruncates),
            self::INSERT()
        ));
    }
```

I stopped searching at `vendor/phpunit/dbunit/PHPUnit/Extensions/Database/Operation/Truncate.php`

May be simpler way would be to create `tearDown()` in postgresqlTest class and reset it there. 

For any search it is good to create a local test env and remove any '@' from database class, ex `@pg_execute()` then you can see any warning.
```
git clone ...
cd project
composer install
JTEST_DATABASE_POSTGRESQL_DSN='host=localhost;port=5432;dbname=[your_db];user=[your_user];pass=[your_pass]' ./vendor/phpunit/phpunit/phpunit

photodude:
Thanks @csthomas that's probably the best explanation of what's going on in the PostgreSQL auto-increment sequences and conditions that are problematic. 

There is definitely something still bugged considering the issues in #53 when we add the PDO version of PgsqlDriver tests back in by uncommenting JTEST_DATABASE_PGSQL_DSN in phpunit.travis.xml

PR #78, opened 2017-01-29
Prepared statement support for SQL Server

mbabker:
### Summary of Changes

This completes having all of the package's drivers support prepared statements by adding support for the driver it was missing in; SQL Server.

### Testing Instructions

Use the `testExecutePreparedStatement` case as an example of how to bind variables for use in a prepared query.

N/A

### DIFF
--- a/src/Sqlsrv/SqlsrvDriver.php
+++ b/src/Sqlsrv/SqlsrvDriver.php
@@ -8,9 +8,12 @@
 
 namespace Joomla\Database\Sqlsrv;
 
+use Joomla\Database\DatabaseQuery;
 use Joomla\Database\Exception\ConnectionFailureException;
 use Joomla\Database\Exception\ExecutionFailureException;
 use Joomla\Database\Exception\UnsupportedAdapterException;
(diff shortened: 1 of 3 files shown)


photodude:
I suggest we kill the composer cache in the .appveyor.yml file I think that is the root of the 
`"./composer.lock" does not contain valid JSON` failure

mbabker:
@wilsonge Review/Merge this so we can finally tag a release here.

src/Sqlsrv/SqlsrvDriver.php
```diff
@@ -604,18 +607,33 @@ public function execute()
...
 		}
-		else
+
+		$params = array();
+
+		// Bind the variables:
+		if ($this->sql instanceof PreparableInterface)
 		{
-			$array = array();
+			$bounded =& $this->sql->getBounded();
```
wilsonge: What's the reason for this being via reference?

mbabker: To be perfectly honest, not a clue.  It's how the other implementations do it though, so I assume there was a reason at some point.

PR #84, opened 2017-03-13
Fix the loading of the table columns

roland-d:
### Summary of Changes
The second argument was missing so the fields were loaded as type only but the code is expecting the full information. Setting the second argument to false, loads all details and the test on line 132 will not throw notices because $column is now an object.

### Testing Instructions
Try to import the XML schema for a database table for an existing table. There will be several notices Notice: Trying to get property of non-object in D:\wamp64\www\joomla-cms\libraries\joomla\database\importer\mysqli.php on line 148

Apply the patch

The import now runs without any notices.

None

### DIFF
--- a/src/Mysqli/MysqliImporter.php
+++ b/src/Mysqli/MysqliImporter.php
@@ -110,7 +110,7 @@ class MysqliImporter extends DatabaseImporter
 	protected function getAlterTableSql(\SimpleXMLElement $structure)
 	{
 		$table = $this->getRealTableName($structure['name']);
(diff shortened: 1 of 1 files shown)


roland-d:
@mbabker Can we get this into 3.7?

roland-d:
@mbabker Any news on this? This bug breaks the importer.

mbabker:
Well, even when merged here it would still have to be patched separate in the CMS because the CMS isn't yet using this code.  So to get it in 3.7 you'll need to patch there.

PR #88, opened 2017-05-30
Events, Monitoring, Remove Debug

mbabker:
### Summary of Changes

- Adds event dispatching for post-connect and post-disconnect actions
- Adds a `QueryMonitorInterface` (roughly based on Doctrine's `SQLLogger` interface) which can be used as a means for monitoring queries (logging, profiling, etc.)
- Removes the debug attribute from the driver, it was only used to toggle whether queries should be logged
- The driver no longer implements PSR-3, queries should be logged through a monitor

- Need proper deprecation for removed functionality
- Document new integration points


wilsonge:
I think that `$this->executed` result should probably be piped into the `stopQuery` method. But after much discussion at the sprint I don't feel strongly enough to override you on it. So merging as is

PR #90, opened 2017-06-20
missed getConnectionCollation()

alikon:
Pull Request for Issue #16789 .

### Summary of Changes
added the missed `getConnectionCollation()` method

### Testing Instructions
Access to System -> System Information

### Expected result
NO error

### Actual result
Call to undefined method Joomla\Database\Mysqli\MysqliDriver::getConnectionCollation()


mbabker:
We need the method in the base driver class too - https://github.com/joomla/joomla-cms/blob/staging/libraries/joomla/database/driver.php#L934 - and probably need to check to make sure if any of the other drivers have this method (I would assume PDO MySQL needs one too).

@wilsonge - Whenever we merge this up to the 2.0 branch, add it to the interface as well

alikon:
thanks

PR #94, opened 2017-06-24
[RFC mysql] - set proper nullDate  depending on SQL_MODE

alikon:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/16788

### Summary of Changes
trivial check for null date

set `$this->nullDate='1000-01-01 00:00:00'; `
if `NO_ZERO_DATE` sql_mode enabled

### DIFF
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -409,7 +409,7 @@ class MysqliDriver extends DatabaseDriver
 	 * @return  string
 	 *
 	 * @since   __DEPLOY_VERSION__
-	*/
+	 */
 	public function getNullDate()
 	{
 		$this->connect();
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -408,7 +408,7 @@ class MysqliDriver extends DatabaseDriver
 	 *
 	 * @return  string
 	 *
-	 * @since    __DEPLOY_VERSION__
+	 * @since   __DEPLOY_VERSION__
 	*/
 	public function getNullDate()
 	{
@@ -418,7 +418,7 @@ class MysqliDriver extends DatabaseDriver
 
 		if (strpos($mode, 'NO_ZERO_DATE') !== false)
 		{
-			$this->nullDate='1000-01-01 00:00:00';
+			$this->nullDate = '1000-01-01 00:00:00';
 		}
 
 		return $this->nullDate;
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -404,6 +404,27 @@ class MysqliDriver extends DatabaseDriver
 	}
 
 	/**
+	 * Get the null or zero representation of a timestamp for the database driver.
+	 *
+	 * @return  string
+	 *
+	 * @since    __DEPLOY_VERSION__
+	*/
+	public function getNullDate()
+	{
+		$this->connect();
+
+		$mode = $this->setQuery('SELECT @@SESSION.sql_mode;')->loadResult();
+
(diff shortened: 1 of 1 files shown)


andrepereiradasilva:
@alikon what happens if a user has a database will old zero date values `0000-00-00 00:00:00` and upgrades mysql now having the new values `1000-01-01 00:00:00`?

is it not possible to just use NULL on cms 4.x?

alikon:
if i have understand quite a bit the problem 
it should arise only when we insert/udpate a  "pseudo-null"   datetime/timestamp
and we are forcing `0000-00-00 00:00:00`
with this pr just returns the "correct" null date (correct for the SESSION sql_mode)

other cms issues still need to be solved ....

andrepereiradasilva:
making the nulldate depends on the databse driver in one thing (that i actually think should be NULL also), but depending on the MySQL version could lead to several issues.

Why not just use `NULL` in mysql and postgresql in the CMS when there is no date? I'm no db expert but do you see any issue with that?

Considering the amount of date fields in the cms db the db size would have shrink a little :) probably also good for performance

richard67:
When using NULL, calculations e.g. of date differences where 1 of the dates is NULL also will result in NULL. Thats's correct, but I have no idea if it could cause issues in Joomla or the framework.
Using NULL when something is not set is the correct way, so we really should check if that is possible.
Changing default values of columns in Joomla from '0000-00-00 00:00:00' to NULL can be safely done with schema update sql scripts.
The function provided by this PR is maybe useful, but maybe it should check the db version and depending on this either return the old or the new null date? I don't know, it depends on the purpose of that function, i.e. where and how it will be used.

alikon:
i don't think is a cms job doing some kind of database migrations task
the only cms task should be work on trusted settings

richard67:
Does that mean you wanna move the utf8mb4 conversio to the framework? ;-)

andrepereiradasilva:
> Changing default values of columns in Joomla from '0000-00-00 00:00:00' to NULL can be safely done with schema update sql scripts.

do you mean something like
```sql
UPDATE yyyy_table
SET zzzz = NULL
WHERE zzzz IN (
  -- MySQL
  '0000-00-00',
  '0000-00-00 00:00:00',
  '1000-01-01',
  '1000-01-01 00:00:00',
  -- PostgreSQL
  '1970-01-01',
  '1970-01-01 00:00:00',
  -- MS SQL
  '1900-01-01',
  '1900-01-01 00:00:00'
)
```
 for all joomla cms db fields?

richard67:
@andrepereiradasilva I was talking about the schema, not the values:
ALTER TABLE gaga MODIFY ... DEFAULT NULL;
It has to be done for all columns of type datetime in a schema update. This is safely checked by the schema check. In the same sql you can also change the values, but this would not be checked by the checker and so not fixed by the fixer when the schema update sql has nt run for some reason, e.g. because silly people still think they can update Joomla just by uploading the files and not even know that it also needs to do something with the database, as I could see in many support forum issues.

richard67:
But that's a cms discussion, and here is the framework.

andrepereiradasilva:
right

mbabker:
Presumably this also needs to be done in the PDO MySQL adapter too.

Also, we should do this in a way where every use of `getNullDate` on MySQL doesn't trigger a query.

alikon:
do you mean something like 

```php
public function getNullDate()
{
	if (in_array($this->options['sqlModes'], 'NO_ZERO_DATE') !== false)
	{
		$this->nullDate = '1000-01-01 00:00:00';
	}

	return $this->nullDate;
}
```
instead ?

alikon:
close in favor of https://github.com/joomla-framework/database/pull/120

PR #95, opened 2017-06-26
Improvements #1

frankmayer:
### Summary of Changes

- Remove callable calls from loop termination conditions\
- Use short syntax for applied operations

### Testing Instructions

Code review only

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -379,7 +379,7 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 
 						if ($comment && $start < $i)
 						{
-							$query = $query . substr($sql, $start, ($i - $start));
+							$query .= substr($sql, $start, ($i - $start));
 						}
 					}
 				}
@@ -394,7 +394,7 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 			{
 				if ($start <= $i)
 				{
-					$query = $query . substr($sql, $start, ($i - $start + 1));
+					$query .= substr($sql, $start, ($i - $start + 1));
 				}
 
 				$query = trim($query);
@@ -403,7 +403,7 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 				{
 					if (($i == $end - 1) && ($current != ';'))
 					{
-						$query = $query . ';';
+						$query .= ';';
 					}
 
 					$queries[] = $query;
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -703,7 +703,7 @@ class MysqliDriver extends DatabaseDriver implements UTF8MB4SupportInterface
 					$bindParams = array();
 					$bindParams[] = &$typeString;
 
-					for ($i = 0; $i < count($params); $i++)
+					for ($i = 0, $iMax = count($params); $i < $iMax; $i++)
 					{
 						$bindParams[] = &$params[$i];
 					}
--- a/src/Pgsql/PgsqlDriver.php
+++ b/src/Pgsql/PgsqlDriver.php
@@ -814,7 +814,7 @@ class PgsqlDriver extends PdoDriver
 			{
 				$sql = explode('currval', $sql);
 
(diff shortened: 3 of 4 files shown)


mbabker:
Closing in favor of #96

It's easier for us to take PRs on master then merge up to the 2.0-dev branch.  So you only need to do changes on master and we'll handle the rest (unless what you're changing is explicitly only in 2.0-dev).

frankmayer:
Ah, ok. wasn't sure & did on both.

frankmayer:
@mbabker maybe it would be good to have this in the readme ;)

mbabker:
Instead of updating 40 READMEs, https://framework.joomla.org/contribute was created to help with that.  So at some point we probably need to make sure all the READMEs link to it (the CONTRIBUTING.md files should already) and if we need any text updates on that main page https://github.com/joomla/framework.joomla.org/blob/master/templates/contribute.twig is the place to do it.

frankmayer:
@mbabker Thanks for the info.

Also, I was about to do PRs which require PHP 5.4 and up (like modern array syntax and other such things). That would be the right case for 2.x, right?

mbabker:
Right.

Just as a frame of reference, for the majority of the repos `master` is 1.x right now and `2.0-dev` for 2.0.  There are a small number of exceptions (we'll get better with this consistency stuff later):

- The Crypt and Form packages use `master` for 2.0 and a `1.x-dev` branch for 1.x (this is the way it really should be, but hindsight is 20/20)
- The Log package has no 2.0 branch as we've deprecated it at the Framework level
- The Renderer package doesn't have a 1.x version, it will only be a part of Framework 2.0 so its `master` branch corresponds to that

For Framework PRs, we generally just go by whether CI passes and a code review, so things can move much quicker in these repos than they do in the CMS repo.

frankmayer:
@mbabker thank you for this in-depth information. If you want, I could put this inside the https://framework.joomla.org/contribute page.
One last thing, just for clarity. Will Framework 2.0 be PHP 5.6+ or will it have to support PHP 5.4, 5.5?

mbabker:
Framework 2.0 and CMS 4.0 are both PHP 5.5.9+ right now, with the option to revisit that before we reach the CMS' beta phase.

mbabker:
And yes, improving the contribute page would be massively helpful.  Just one less thing on my mile long Framework documentation TODO list 😆

frankmayer:
OK! :smile: 
However PHP 5.5 is EOL for almost a year now (http://php.net/supported-versions.php). Shouldn't the minimum be 5.6 now? Why still 5.5?

mbabker:
That's what we started at when we started the 4.0 journey as more of a conservative thing (also a majority of the PHP ecosystem had at that point standardized on 5.5.9 for their releases, a lot of projects coming up to new major releases are evaluating their minimums as well).  But we also made it pretty clear that we'd most likely revisit it when we got closer to a beta release, and using [our stats](https://developer.joomla.org/about/stats.html) as support odds are 5.5 support will disappear.

frankmayer:
Hm, I see.
Now, I don't know when this journey started :) but PHP 5.5 is gone and no one should use it anyways. Why not make a decision right now and open the doors to even more progress, by making use of the 5.6 goodies, instead of having to reevaluate code later again for no "logical" reason at all?

mbabker:
We just hadn't reached a discussion point yet is all.  It's something I want to open the door on sooner than later, but I'm also aware there are other priorities and that there aren't major language features that need us to drop everything and have that discussion yet.  So I'm comfortable with letting it ride for another couple of months right now, especially as we won't have a beta until at the earliest (and this is rushing it IMO) around the 3.8 stable release (still roughly 3 months away).

Personally, I think once we've got a lot of the namespacing and re-integration of the Framework tasks knocked out on the CMS repo for 3.8/4.0, that can open the door to having that talk again.  And I think we're really close on that.

frankmayer:
OK, awaiting the big decision, then :) hopefully sooner than later. Thanks

mbabker:
FWIW I just did [a round of changes](https://github.com/joomla/framework.joomla.org/commit/0bf9e4ebdfb7c69b3175fb248410f8fd2cc6c642) for the contribute page.  Hopefully that helps folks out some.

PR #98, opened 2017-06-26
Improvements #3

frankmayer:
### Summary of Changes
- Fix misordered modifiers
- Remove unnecessary parentheses
- Remove one-time-use variables
- Remove useless return

### Testing Instructions
Code review only   (w=1 url parameter is probably a good idea, as I have also formatted some code)


andrepereiradasilva:
seems fine to me. thanks

PR #101, opened 2017-08-05
[2.0] Use Symfony's OptionsResolver for database options

mbabker:
As a way to more explicitly be aware of and configure the various options used in the database drivers, we can simplify things down a bit and use Symfony's OptionsResolver component to define the various options, the allowed option types/values, and set default values.  This also removes the consistent need for `isset` checks as all options should be defined going forward.

The one catch here though is the resolver is by design a bit strict about what values it will accept, and this will make it so that you cannot inject options into the constructor that the resolver's definition does not know for a driver class.


mbabker:
`¯\_(ツ)_/¯`

PR #104, opened 2017-11-28
Fix sqlsrv syntax for query SELECT with DISTINCT and LIMIT

csthomas:
### Summary of Changes

This is changes applied in J3.8.2 from https://github.com/joomla/joomla-cms/pull/18315
I added new missing methods and variables to `SqlsrvQuery` class.

Ping @wilsonge 

Note:
I'm not sure, may be would be better to set for `$limit` default value `null` instead `0`?

### DIFF
--- a/Tests/Sqlsrv/SqlsrvIteratorTest.php
+++ b/Tests/Sqlsrv/SqlsrvIteratorTest.php
@@ -51,8 +51,8 @@ class SqlsrvIteratorTest extends SqlsrvCase
 				2,
 				0,
 				array(
-					(object) array('title' => 'Testing', 'RowNumber' => '1'),
-					(object) array('title' => 'Testing2', 'RowNumber' => '2')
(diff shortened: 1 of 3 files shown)


alikon:
I would be even more further, SQL Server on linux start from 2017 iirc ...

csthomas:
The list of versions https://sqlserverbuilds.blogspot.sg/

mbabker:
> still fighting with https://docs.microsoft.com/en-us/sql/linux/quickstart-install-connect-docker?view=sql-server-linux-2017 so unable to do a real test but OK from a code review

Luckily Travis is spun up now to run that exact Docker image so the unit/functional tests here are actually even more useful than before.

alikon:
then green light on this

alikon:
sorry for cross posting any chance  on this one  https://github.com/joomla/joomla-cms/pull/19707 ?

PR #111, opened 2018-03-18
[2.0] Have the base QueryInterface extend PrepareableInterface, require prepared statement support always

mbabker:
### Summary of Changes

We're now at a point where all drivers support prepared statements.  However, this is still an optional feature and it is possible for query classes to be built without the support.  We should discourage that behavior, therefore I suggest that we have our QueryInterface extend the PrepareableInterface and in effect mandate that Query objects (and drivers) support prepared statements.

### Testing Instructions

N/A, no API changes

Note interface requirements

### DIFF
--- a/Tests/Mock/Query.php
+++ b/Tests/Mock/Query.php
@@ -13,4 +13,104 @@ namespace Joomla\Database\Tests\Mock;
  */
 class Query extends \Joomla\Database\DatabaseQuery
 {
+	/**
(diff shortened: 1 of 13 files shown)


mbabker:
@wilsonge Added docs covering a decent chunk of the upgrade path thus far.

wilsonge:
Star! LGTM

PR #114, opened 2018-04-01
[2.0] Introduce Statement Objects

mbabker:
### Summary of Changes

The manner in which prepared statements are implemented right now causes PHP with MySQL support compiled against `libmysql` to fail over miserably.  This is because our implementation relies on `mysqli_stmt::get_result()` to get a result set cursor, which is only available when `mysqlnd` is in use.

This causes deeper issues in our existing API because the API relies on having this result set cursor to be able to do any other function, and to the best of my knowledge there is no other way to get a result set cursor when using prepared statements with MySQLi.  This means to make things work there's some major underlying architectural changes needed, and that's what this PR provides.

First, the B/C breaks:

- `DatabaseInterface::execute()` now ONLY returns boolean, it will no longer return a result set cursor
- `DatabaseDriver` subclasses MUST implement a new `prepareStatement` method which creates a new `StatementInterface` implementation for that driver and assigns it to the driver's `$statement` property
- The first argument of `DatabaseIterator` is changed from a result set cursor to a `StatementInterface` object, this should be passed after the query is executed (`DatabaseDriver::getIterator()` accounts for this)
- Methods of `DatabaseDriver` which accepted a result set cursor as an argument have had this argument removed, they will now only work with the statement assigned in the `$statement` property

Next, the API design model:

Essentially, this is the same thing we are already doing for PDO, and the same way Doctrine handles statements and result sets.  The key here is our `StatementInterface` is a stub of `PDOStatement` exposing what we need to keep our API functional for our non-PDO drivers.  Also to keep the hard dependency on PDO from creeping in, its constants that are used as method arguments are enumerated through a couple of marker classes.

This effectively moves a lot of the connection specific logic out of the `DatabaseDriver` class chain and into the Statement objects, so there's some code moved into the base class and removed from its subclasses now since the API is a bit more consistent.

Also, because of the changes needed to make the iterators work, there is now no longer a need to have `DatabaseIterator` subclasses and the `DatabaseFactory` has been adjusted so that if a driver specific subclass doesn't exist the base class is returned.

### Testing Instructions

Use API as normal.  This really needs tested on an environment where PHP is built using `libmysql`, every environment I have access to uses `mysqlnd` and doesn't hit the root problem causing this refactoring to be required (including AppVeyor and Travis-CI).  The tests are all green so with any luck just running PHPUnit on that kind of environment will answer pretty quickly whether the root issue is fixed.

As of this writing AppVeyor is still catching up with all of my changes, so I have no idea what state SQL Server is in for the moment but I'll keep an eye on it.

See above.


nikosdion:
So, hi, I'm here with @wilsonge and I tested the CMS PR 20055 which he tells me is the same thing as this PR. So here are my test results.

A. PHP 7.0 without mysqlnd.
A.1. Without the patch. I can't see the mysqli driver at all.
A.2. With the patch. I can see the mysqli driver and install Joomla.

B. PHP 7.1 with mysqlnd.
B.1. Without the patch. I can see the mysqli driver BUT the installation fails. The tables are not created, there is no error thrown during that bit but the application crashes as soon as it tries to save a session since the default handler is database and we, uh, don't have one :)
B.2. With the patch. I can see the mysqli driver and install Joomla.

I understand that this makes George lose a bet. Sorry, not sorry :p

mbabker:
Thanks a ton!!!

As soon as I get the CI failure for the SQL Server driver fixed this is being merged then going downstream to the CMS repo.

mbabker:
This is passing locally, Travis CI for SQL Server needs some more tweaks, will work on that later.

PR #118, opened 2018-04-24
Add a method DatabaseQuery::isNullDatetime()

csthomas:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/16788

### Summary of Changes

Add a new method `DatabaseQuery::isNullDatetime($column)` to allow to compare column with more than one ZERO/NULL value.

Another ideas:
- add `9999-12-31 23:59:59` to nullDatetimes .
- add additional methods like `isNullOrLowerDatetime()` or `isNullOrHigherDatetime()`, maybe without text 'NullOr'.

An example for mysql:
```php
$db->getQuery(true)
    ->select('*')
    ->from($db->quoteName('#__content'))
    ->where('publish_up < ' . $db->quote('2017-01-01 00:00:00') 
        . ' OR ' . $query->isNullDatetime('publish_up'));
```
will be 

```sql
SELECT * 
FROM `#__content` 
WHERE publish_up < '2017-01-01 00:00:00' 
OR (publish_up IN ('0000-00-00 00:00:00', '1000-01-01 00:00:00') OR publish_up IS NULL)
```

### Testing Instructions
At now only code review.

### Note
~~Method `DatabaseQuery::isNullDatetime($column)` automatically calls `$db->quoteName($column)` inside.~~ - reverted.

### DIFF
--- a/Tests/DatabaseQueryTest.php
+++ b/Tests/DatabaseQueryTest.php
@@ -1297,6 +1297,23 @@ class DatabaseQueryTest extends TestCase
 	}
 
 	/**
+	 * Tests the isNullDatetime method.
+	 *
+	 * @return  void
+	 *
+	 * @covers     \Joomla\Database\DatabaseQuery::isNullDatetime
+	 * @since      __DEPLOY_VERSION__
+	 */
+	public function testIsNullDatetime()
+	{
+		$this->assertThat(
+			$this->instance->isNullDatetime('publish_up'),
+			$this->equalTo('`publish_up` IS NULL'),
+			'Test isNullDatetime failed.'
+		);
+	}
+
+	/**
 	 * Tests the \Joomla\Database\DatabaseQuery::nullDate method for an expected exception.
 	 *
 	 * @return  void
--- a/Tests/DatabaseQueryTest.php
+++ b/Tests/DatabaseQueryTest.php
@@ -1308,7 +1308,7 @@ class DatabaseQueryTest extends TestCase
 	{
 		$this->assertThat(
 			$this->instance->isNullDatetime('publish_up'),
-			$this->equalTo('`publish_up` IS NULL'),
+			$this->equalTo('publish_up IS NULL'),
 			'Test isNullDatetime failed.'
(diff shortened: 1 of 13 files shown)


src/DatabaseDriver.php
```diff
@@ -117,6 +117,14 @@
 	 */
 	protected $nullDate;
 
+	/**
+	 * The list of zero or null representation of a datetime for the database driver.
+	 *
+	 * @var    string
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $nullDatetimes;
```
mbabker: I'd say this should always be an array instead of a string with comma separated values.

src/DatabaseDriver.php
```diff
@@ -1148,6 +1156,22 @@ public function isMinimumVersion()
...
+	/**
+	 * Generate a SQL statement to check if column represents a zero or null datetime.
+	 *
+	 * @param   string  $column  A column name.
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function isNullDatetime($column)
```
mbabker: If it's building SQL this seems like it should be in the `DatabaseQuery` class.

src/DatabaseDriver.php
csthomas: I thought so myself, until I found that this variable is read many times and sets only once. I used a trick. I will not oppose.

src/DatabaseDriver.php
csthomas: Should I move `protected $nullDatetimeList = [];` to `DatabaseQuery` too? IMO yes.

csthomas:
It is ready to the next code review.

csthomas:
To be more consistency with other methods the last commit removes  `$column = $this->db->quoteName($column);` from `isNullDatetime($column)`.

csthomas:
@mbabker Any thoughts about this PR? Should I change something?

csthomas:
Thanks

PR #120, opened 2018-05-01
mysql - null date from sql mode

alikon:
redo  of https://github.com/joomla-framework/database/pull/94
https://github.com/joomla/joomla-cms/issues/16788#issuecomment-385579869

### Summary of Changes
set $this->nullDate='1000-01-01 00:00:00';
depending on wich sql mode is enabled `NO_ZERO_DATE `

### Testing Instructions
change the current hard coded setting in https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/vendor/joomla/database/src/Mysqli/MysqliDriver.php#L94

### DIFF
--- a/Tests/Mysqli/MysqliDriverTest.php
+++ b/Tests/Mysqli/MysqliDriverTest.php
@@ -766,4 +766,25 @@ class MysqliDriverTest extends MysqliCase
 	{
 		$this->markTestIncomplete('This test has not been implemented yet.');
 	}
+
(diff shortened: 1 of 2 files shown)


mbabker:
Can you add a test case to https://github.com/joomla-framework/database/blob/2.0-dev/Tests/Mysqli/MysqliDriverTest.php please?  Otherwise this seems good to me.

PR #124, opened 2018-05-10
[2.0] Refactor SqlsrvQuery::processList method

csthomas:
### Summary of Changes

Simpler OFFSET/LIMIT clause for sql server.
This PR requires SQL Server 2012.

Because `OFSET/FETCH` has to follow by `ORDER BY` I added a comment `/*ORDER BY (SELECT 0)*/` if ORDER BY clause does not exists in query. 

Later I can easy uncomment it if  `OFFSET/FETCH` is added.

### Testing Instructions
Code review.
It would be good to do a real test.

```php
require 'vendor/autoload.php';
$db = Joomla\Database\DatabaseDriver::getInstance(['driver' => 'sqlsrv', 'database' => '', 'user' => '', 'password' => '', 'prefix' => 'j38_']);
print_r($db->setQuery($db->getQuery(1)->select('id')->from('j38_modules'), 0, 3)->loadAssocList());
print_r($db->setQuery($db->getQuery(1)->select('id')->from('j38_modules'), 1, 3)->loadAssocList());
print_r($db->setQuery($db->getQuery(1)->select('id')->from('j38_modules'), 1)->loadAssocList());
print_r($db->setQuery($db->getQuery(1)->select('id')->from('j38_modules'))->loadAssocList());
```

**Known issue:**
When you use one of `SELECT DISTINCT` or `GROUP BY` and you want to use `OFFSET` or `OFFSET/LIMIT` then you have to add `ORDER BY` clause.

To workaround that I may replace `/*ORDER BY (SELECT 0)*/` by `ORDER BY 1` but IMO this trick should not be added.

See https://stackoverflow.com/questions/34765786/nhibernate-order-by-current-timestamp

### DIFF
--- a/Tests/Sqlsrv/SqlsrvIteratorTest.php
+++ b/Tests/Sqlsrv/SqlsrvIteratorTest.php
@@ -66,8 +66,8 @@ class SqlsrvIteratorTest extends SqlsrvCase
 				20,
 				2,
 				array(
-					(object) array('title' => 'Testing3', 'RowNumber' => '3'),
-					(object) array('title' => 'Testing4', 'RowNumber' => '4')
+					(object) array('title' => 'Testing3'),
+					(object) array('title' => 'Testing4')
 				),
 				null
 			),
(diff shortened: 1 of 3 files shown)


csthomas:
I will change it.

PR #127, opened 2018-05-14
Escaping float value should be non-locale aware

csthomas:
### Summary of Changes

When I run unit tests locally the result always failed because of my locale (pl_PL).
In Polish, numbers are displayed with the comma as a decimal point.

If I use `$query->where('column = ' . $db->quote(3.14)); // <- input a float value`
then the query instead of displays `... WHERE column = 3.14` shows `... WHERE column = 3,14` which is incorrect.

See https://github.com/matomo-org/matomo/issues/6435

### Testing Instructions
Code review.
Unit tests pass.

No

### DIFF
--- a/Tests/DatabaseDriverTest.php
+++ b/Tests/DatabaseDriverTest.php
@@ -354,7 +354,7 @@ SQL
 	 */
 	public function testQuoteFloat()
 	{
-		// This also call `escape()` method from nosqldriver which is locate aware
+		// This call `escape()` method from nosqldriver, which is locale aware
 		$this->assertSame(
 			// Below line may generate "'-3.14-'" or "'-3,14-'"
 			"'-" . 3.14 . "-'",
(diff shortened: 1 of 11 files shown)


mbabker:
Not entirely sure what to think about this yet.  I feel like this could create problems if the database is somehow in a state where numbers are stored in the "3,14" format (if that's even possible).  While we encourage using `utf8` aware locales/collations, I'm not aware of anything in our code forcing it or explicitly prohibiting use of language specific locales/collations.

csthomas:
Take a look at 
* https://stackoverflow.com/questions/8669212/change-decimal-separator-in-mysql
* https://stackoverflow.com/questions/6627239/insert-non-english-decimal-points-in-mysql/6627551#6627551

mbabker:
Oh, well, in that case...

Issue #112, opened 2018-03-20
Need consistent syntax for parameterized query support


mbabker:
Starting with the 1.5.0 tag, all drivers have support for prepared statements and parameterized queries, and in 2.0 this will be a hard requirement because of the new `QueryInterface` dictating it.  However, the current implementation is dependent on the syntax(es) supported by the underlying driver, meaning at the moment it isn't possible to actually write "properly" abstract queries with parameterized query support (unless an application is only supporting PDO + MySQLi)

- Question mark syntax (`INSERT INTO foo (col1, col2) VALUES (?, ?)`) is supported by PDO, MySQLi, and SQL Server
- Named parameter syntax (`INSERT INTO foo (col1, col2) VALUES (:col1, :col2)`) is supported by PDO
- "PHP variable" syntax (`INSERT INTO foo (col1, col2) VALUES ($1, $2)`) is supported by PostgreSQL (`ext/pgsql`, separate from the PDO driver)

So in essence we need some kind of SQL parsing handler in the `DatabaseQuery` class.  My thought was to have it support the question mark syntax as that one is already supported by most drivers and we only need a PostgreSQL implementation.  While discussing in chat @nibra suggested we standardize around the named parameter syntax since it creates an easier to read and parse string.

Whatever the end result, we do need a solution that allows users of our API to not be concerned with these driver specific details (it's OK for now if you're writing single database applications, but because the CMS is using this package in 4.0, it becomes a problem).

IsaacWG:
I much prefer the named parameter syntax myself. I agree with it being easier to read. I do NOT like the "PHP Varible" syntax as that could get confused with actual PHP varibles and would become a problem if a developer uses a doublequoted string instead of single quotes.

mbabker:
To be honest I wouldn't even consider that an option, but since that's the syntax `ext/pgsql` uses I put it in here in part to demonstrate what all exists from a practical perspective and to demonstrate why we need the parser layer.

nibra:
IMHO we should prefer the named parameter syntax, because
a) it is easier to read (fx. find an unwanted swap of parameters) and
b) it is easier to parse.
The specific driver can then convert that internally to any bespoke syntax, if needed.

drmmr763:
I agree with @nibra - named parameters are generally easier to debug. Especially because moving to named params makes "copy / paste and run this sql directly" a pain for devs not used to it. 

A few things i've ran into writing some libraries are:
[...]

PR #129, opened 2018-05-15
Add named prepared parameters to mysqli driver

HLeithner:
We add a translation function for named parameters to positioned parameters. This is similar like it is done by php in the pdo driver.  #112

### Summary of Changes
Add prepareParameterKeyMapping function
Use new parameters
Add test

### Testing Instructions
Try named parameters with mysqli driver
ex.
```php
$client_id = 1;
$query->select('*')->from($db->quoteName('#__session'));
$query->where($db->quoteName('client_id') . ' = :client_id');
$query->bind(':client_id', $client_id, ParameterType::INTEGER);

$db->setQuery($query);
$list = $db->loadObjectList();
```

### DIFF
--- a/src/Mysqli/MysqliStatement.php
+++ b/src/Mysqli/MysqliStatement.php
@@ -307,6 +307,8 @@ class MysqliStatement implements StatementInterface
 			{
 				$params[$this->parameterKeyMapping[$key]] =& $value;
 			}
+
+			ksort($params);
 		}
 		else
 		{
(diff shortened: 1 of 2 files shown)


src/Mysqli/MysqliStatement.php
```diff
@@ -115,22 +123,145 @@ class MysqliStatement implements StatementInterface
 	 */
 	public function __construct(\mysqli $connection, string $query)
 	{
-		$this->connection = $connection;
-		$this->query      = $query;
+		$this->connection   = $connection;
+		$this->query        = $query;
+		$this->bindedValues = null;
```
csthomas: The above line is redundant.

PR #141, opened 2018-11-03
Replace serialize/unserialize with simple cloning

csthomas:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/22685

### Summary of Changes
Use object cloning instead of serialization.

### Testing Instructions
See unit test.

### Note
`DatabaseQuery/QueryElement` object should not have any multidimensional array with objects.
Each member of the `DatabaseQuery/QueryElement` object should take care of its own `__clone()` method, (except `DatabaseQuery::db`).

### DIFF
--- a/Tests/DatabaseQueryTest.php
+++ b/Tests/DatabaseQueryTest.php
@@ -1987,6 +1987,32 @@ class DatabaseQueryTest extends TestCase
 	}
 
 	/**
+	 * Tests the \Joomla\Database\DatabaseQuery::__clone method properly clones nested objects.
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function test__clone_nested_query()
+	{
+		/**
+		 * Add a custom closure to test the cloning of nested objects.
+		 *
+		 * DatabaseQuery->QueryElement->DatabaseQuery
+		 * $query       ->merge[0]    ->elements[0]
+		 */
+		$this->dbo->func = function() {};
+
+		$query = $this->instance->select('id')->from('#__content');
+		$query->union(clone $query);
+
+		// Real test for exception: Serialization of 'Closure' is not allowed
+		$query2 = clone $query;
+
+		$this->assertFalse($query->merge[0]->getElements()[0] === $query2->merge[0]->getElements()[0]);
+	}
+
+	/**
 	 * Tests the \Joomla\Database\DatabaseQuery::union method.
 	 *
 	 * @return  void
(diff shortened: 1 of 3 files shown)


ggppdk:
I have been using this last days, so far looks good

Regarding this comment

> DatabaseQuery/QueryElement object should not have any multidimensional array with objects.
Each member of the DatabaseQuery/QueryElement object should take care of its own __clone() method, (except DatabaseQuery::db).

Maybe something out of the above should be added to the description of 

```php
	/**
	 * Method to provide deep copy support to nested objects and arrays when cloning.
	 *
	 * @return  void
	 *
	 * @since   1.0
	 */
	public function __clone()
	{
```

also is the "deep" copy comment still valid ? i mean now the "cloning" is 2 levels deep, right ?

csthomas:
>> DatabaseQuery/QueryElement object should not have any multidimensional array with objects.
>> Each member of the DatabaseQuery/QueryElement object should take care of its own __clone() method, (except DatabaseQuery::db).
>
> Maybe something out of the above should be added to the description of

OK.

Each `DatabaseQuery / DatabaseElement` child object requires its own `__clone()` method.

If you add a new object, ex (`DatabaseAsset`) as `DatabaseQuery` variable, then you have to guarantee its own `__clone() `method, if it is required.

Notice: The `DatabaseDriver` object should be never cloned.

I have not added any support for cloning object in multidimensional array.
I do not see any reason for that.  There is no object in multidimensional array in the current code.

In general, a multidimensional array of immutable variables (integer, string, float, ...) is safe.

csthomas:
I have added a better description.

csthomas:
> Does that mean if you subclass the QueryElement and/or DatabaseQuery classes you need your own __clone() implementation?

No, but if (edge case) you will use multidimensional array of objects in above subclasses then you will have to extend `__clone()` method in above subclasses.

> any object that's pushed into the data of these classes should have its own __clone() implementation?

Yes, but only if your object contains mutable variables, means objects.

No, if (data in `YourObject` is immutable):
-  `DatabaseQueryObject -> QueryElementObject -> YourObject -> string`
-  `DatabaseQueryObject -> QueryElementObject -> YourObject -> int`
- `DatabaseQueryObject -> QueryElementObject -> YourObject -> array(int, string, float, boolean)`

Yes, if 
- `DatabaseQueryObject -> QueryElementObject -> YourObject -> stdObject`
- `DatabaseQueryObject -> QueryElementObject -> YourObject -> array(stdObject, ...)`

then you have to implement `YourObject::__clone()` method.

Each child object have to care of the cloning of its own mutable data. 

If you do not do this, the mutable data of a child will not be cloned, but will only be a reference.

I would be thankful if you could add a correct explanation in the description.
It is difficult for me to create a correct description in English.

PR #143, opened 2018-11-06
Add a condition parameter to the join methods

csthomas:
### Summary of Changes

Create a more flexible and clear structure for the joining elements.
Move the sql text `ON` to internal structures.

At the moment this changes are B/C, but I have removed the previous examples from methods documentation.

I have moved several methods to sort them in sequence (`join`,` innerJoin`, `outerJoin`,` leftJoin`, `rightJoin`).
 

### Testing Instructions
Unit tests should be enough

Yes. TO DO

### DIFF
--- a/Tests/DatabaseQueryTest.php
+++ b/Tests/DatabaseQueryTest.php
@@ -1163,6 +1163,14 @@ class DatabaseQueryTest extends TestCase
 			$this->equalTo(TestHelper::getValue($q2, 'join')),
 			'Tests that innerJoin is an alias for join.'
 		);
+
+		$q3 = $this->dbo->getQuery(true)->innerJoin('foo', 'foo.id = bar.id');
+
+		$this->assertThat(
+			(string) TestHelper::getValue($q3, 'join')[0],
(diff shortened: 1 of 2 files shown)


csthomas:
Should I add other methods to QueryInterface: leftJoin, ...?

mbabker:
I purposefully left the convenience shortcut methods off.

csthomas:
OK

ggppdk:
Old syntax said that $conditions could be an array
```php
	 * @param   string  $conditions  A string or array of conditions.
```

but passing array was not possible because $glue was not passed when creating the QueryElement
```php
$this->join[] = new Query\QueryElement(strtoupper($type) . ' JOIN', $conditions, -NO GLUE PASSED-);
```

you would need something like
```php
$t = strtoupper($type) . ' JOIN';
$this->join[] = new Query\QueryElement($t, $conditions, ' ' . $t . ' ');
```

thus you could only pass 1 condition to join(...)
so i guess the new documentation of the parameters is good

PR #145, opened 2018-11-16
Correctly escape quotes in quoteName method

csthomas:
### Summary of Changes

In my previous PR I forgot to quote the delimiter.

See, https://docs.microsoft.com/en-us/sql/t-sql/functions/quotename-transact-sql?view=sql-server-2017

Port of PRs https://github.com/joomla/joomla-cms/pull/13860 and https://github.com/joomla/joomla-cms/pull/13825

### Testing Instructions
Unit tests should pass.
Code review.

No

### DIFF
--- a/Tests/DriverMysqlTest.php
+++ b/Tests/DriverMysqlTest.php
@@ -29,6 +29,22 @@ class DriverMysqlTest extends DatabaseMysqlCase
 	}
 
 	/**
+	 * Data for the testQuoteName test.
+	 *
+	 * @return  array
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function dataTestQuoteName()
+	{
+		return array(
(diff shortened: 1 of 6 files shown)


mbabker:
@csthomas can you revisit this change?  Now that I (finally) tagged a release with it, I'm getting test failures on the [stats server](https://github.com/joomla/statistics-server) repo pointing to this change, see https://travis-ci.org/joomla/statistics-server/builds/505284669 for reference (don't have time this second to check if it's just a bug in the test setup or an actual code issue but figured I'd ping and make sure).

csthomas:
Hi,

At first glance, it looks like there is a problem with (mocked) database object in tests.
In some where, this `$db->nameQuote` has to be defined.

https://github.com/joomla-framework/database/blob/64377b02dba550053adba8a11a6b310731f9929a/src/DatabaseDriver.php#L106

PR #146, opened 2018-11-23
Add a new method `alias()` to DatabaseQuery

csthomas:
### Summary of Changes

Add a new method `$query->alias('alias')` to help you to generate the query string as `(SELECT ... FROM ...) AS alias`.

Note: `$query->alias()` does not call `$db->quoteName()` `implicitly.

### Example
```php
$sub = $db->getQuery(true);
$sub->select('col AS col2')->from('table')->where('a=1');
$query = $db->getQuery(true)->select('col2')->from($sub->alias('alias'));
```

### Testing Instructions
Code review.
Unit tests should pass.

This PR requires https://github.com/joomla-framework/database/pull/141 to work properly.

Yes.

- New method `alias($alias)`.
- Changes in method `from($table)`.

### DIFF
--- a/Tests/DatabaseQueryTest.php
+++ b/Tests/DatabaseQueryTest.php
@@ -136,7 +136,7 @@ class DatabaseQueryTest extends TestCase
 	public function test__toStringFrom_subquery()
 	{
 		$subq = clone $this->instance;
-		$subq->select('col AS col2')->from('table')->where('a=1');
+		$subq->select('col AS col2')->from('table')->where('a=1')->setLimit(1);
 
 		$this->instance->select('col2')->from($subq->alias('alias'));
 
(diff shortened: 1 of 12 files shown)


csthomas:
Can I add these changes to the main branch? I know I will have to remove drastic changes (non B/C) in the `from()` method.

mbabker:
The new method is fine for 1.x (master).  The other changes I'd be cautious about unless you can work out a way to be B/C on it (and we'll probably still need a B/C layer in 2.0 but the signature change would be OK under SemVer).

I'll look more over the weekend at things.

csthomas:
I tried to do it for the master branch, but without  PR https://github.com/joomla-framework/database/pull/141 this won't go. I stay with 2.0-dev. I will add some minor changes only.

csthomas:
Now it should be ready

PR #147, opened 2018-12-07
Add a new class DebugMonitor

csthomas:
### Summary of Changes

To help J4 to display all queries in logs when debug is ON.

The idea on J4 is to add a `DebugMonitor` object to DatabaseDriver instance when JDEBUG is ON.

```php
[...]
$options = [
        'driver'   => $dbtype,
        'host'     => $conf->get('host'),
        'user'     => $conf->get('user'),
        'password' => $conf->get('password'),
        'database' => $conf->get('db'),
        'prefix'   => $conf->get('dbprefix'),
];

if (defined('JDEBUG') && JDEBUG)
{
        $options['monitor'] = new \Joomla\Database\Monitor\DebugMonitor();
}

$db = DatabaseDriver::getInstance($options);
```

Joomla plugin will have an option to get/set debug monitor based on other configuration.

This is a copy/paste of `DebugMonitor` from https://github.com/joomla/joomla-cms/blob/4.0-dev/plugins/system/debug/DebugMonitor.php with a few changes.

I'm not sure about:
- use the `final` keyword in declaration of class or not
- use `private` or `protected` variables

At the moment there is a "final class". 
And "private" variables.

```php
// A way for J4 debug plugin
$monitor = $db->getMonitor();

if (!$log_queries) { 
    $db->setMonitor(null);
}
```

### Testing Instructions
Code review.

No

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1732,7 +1732,7 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	 *
 	 * @since   __DEPLOY_VERSION__
 	 */
-	public function setMonitor($monitor)
+	public function setMonitor(QueryMonitorInterface $monitor = null)
 	{
 		$this->monitor = $monitor;
 
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1728,6 +1728,21 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	}
 
 	/**
+	 * Delete the query monitor.
+	 *
+	 * @return  QueryMonitorInterface|null
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function delMonitor()
+	{
+		$monitor       = $this->monitor;
+		$this->monitor = null;
+
+		return $monitor;
+	}
+
+	/**
 	 * Sets the SQL statement string for later execution.
 	 *
 	 * @param   string|QueryInterface  $query   The SQL statement to set either as a Query object or a string.
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1728,13 +1728,13 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	}
 
 	/**
-	 * Delete the query monitor.
+	 * Remove the query monitor.
 	 *
-	 * @return  QueryMonitorInterface|null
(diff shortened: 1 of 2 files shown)


src/DatabaseDriver.php
```diff
@@ -1727,6 +1727,21 @@ public function setMonitor(QueryMonitorInterface $monitor)
...
 	}
 
+	/**
+	 * Remove the query monitor.
+	 *
+	 * @return  QueryMonitorInterface|null  The removed monitor or null if not set.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function removeMonitor()
```
mbabker: If this is going to be done this should be a setter that allows a null argument.

mbabker:
> I'm not sure about:
> 
> * use the `final` keyword in declaration of class or not
> * use `private` or `protected` variables
> 
> At the moment there is a "final class".
> And "private" variables.

As far as extensibility goes, it is easier to start with a class being final and class member properties being private.  It means the class is not designed to be extended (use the interfaces and build your own custom implementations).  A class without the final keyword is an extension point, makes working with B/C a bit more difficult.  A class member property that is visible in subclasses is an extension point, makes working with B/C a bit more difficult.  So the approach I've taken to code the last few months has basically been "default to least access required" and requiring justification on making things more accessible (i.e. dropping final keyword or moving private to protected).

It makes you really think about your API design and what you want, or don't want, consumers to do with things.

src/DatabaseDriver.php
csthomas: Then I replace this method with `getMonitor()` and allow for null in `setMonitor()`.

csthomas:
If it's worth it, I can add
```diff
diff --git a/src/DatabaseDriver.php b/src/DatabaseDriver.php
index 02d01d7..b8e0b8a 100644
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1734,6 +1734,13 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
         */
        public function setMonitor($monitor)
        {
+               if (!$monitor instanceof QueryMonitorInterface && $monitor !== null)
+               {
+                       throw new \InvalidArgumentException(
+                               sprintf('A monitor must be a %s instance or be null.', QueryMonitorInterface::class)
+                       );
+               }
+
                $this->monitor = $monitor;
 
                return $this;
```

csthomas:
> public function setMonitor(QueryMonitorInterface $monitor = null)

Thanks

PR #149, opened 2018-12-22
Add support for binary data

csthomas:
### Summary of Changes

Add two new methods:
- `$db->quoteBinary()` - required by ms sql server (to explicitly cast a value on INSERT query)
- `$db->decodeBinary()` - required by Postgresql/Pgsql drivers to get the original binary string.

Add a missing class `DatabaseSqliteCase` in order to load `./Stubs/ddl.sql` file and the same `./Stubs/database.xml` file as on other drivers. 
Previously, phpunit loaded SQL/XML data from the additional package 'Joomla\Test'.

### Testing Instructions
Code review.
Unit tests should pass

To work properly with a binary string, we have to quote it using `quoteBinary()`. This method replaces the binary string with hexadecimal text wrapped in the specified format depending on the database driver.

When we want to get value of `BLOB/VARBINARY` then Postgresql/Pgsql database API returns an equivalent of binary string. 

Postgresql (>= 9.0) by default returns hexdecimal text or (<9.0) escaped string.
I changed SQL variable `buffer_output` to always return escaped string and then use `pg_unescape_string` to get the original binary string.

PgSQL returns a stream instead of string. To get the original binary string we have to use `stream_get_contents()`. 

Other databases do not require this additional conversion. For the code to work properly regardless of the database driver, we will use  `$db->decodeBinary($value)`.

### DIFF
--- a/Tests/DatabaseSqliteCase.php
+++ b/Tests/DatabaseSqliteCase.php
@@ -1,0 +1,112 @@
+<?php
+/**
+ * @copyright  Copyright (C) 2005 - 2018 Open Source Matters, Inc. All rights reserved.
+ * @license    GNU General Public License version 2 or later; see LICENSE
+ */
+
+namespace Joomla\Database\Tests;
+
+use Joomla\Database\DatabaseDriver;
+use Joomla\Database\Sqlite\SqliteDriver;
+use Joomla\Test\TestDatabase;
+
+/**
+ * Abstract test case class for MySQLi database testing.
+ *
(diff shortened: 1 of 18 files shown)


mbabker:
I shouldn't review PRs on mobile.  I thought this was targeted to the 2.0 branch, so https://github.com/joomla-framework/database/pull/149/commits/889c2f2f8248ccfc96142c5baa96b6da4968e81c should be reverted if it's going to master.  My bad.

PR #150, opened 2018-12-27
Fix unit tests for mysqli driver

csthomas:
### Summary of Changes

Simple fix in doc

### Testing Instructions
Unit tests should pass on mysqli driver


peter279k:
It seems that the Travis CI build is failed because of `mssql` extension installation failure.

According to the master branch of original repository, this installation has been removed.

And I think this PR is only about fixing comment style.

PR #156, opened 2019-05-07
[FEATURE] add whereNotIn function

f-hamel:
### Summary of Changes
Add function `whereNotIn`.

So now we can use prepared statements for `NOT IN` queries easier.

### Testing Instructions
using something like this:

```php
$groups = array("Editor", "Author");
$db = Factory::getDbo();
$query = $db->getQuery(true)
                                ->select($db->quoteName('title'))
                                ->from($db->quoteName('#__usergroups'))
                                ->whereNotIn($db->quoteName('title'), $currentGroups, ParameterType::STRING);$db->setQuery($query);

$result = $db->loadColumn()
```

`$result` should contain alls groups, except of `Editor` and `Author`

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -1706,6 +1706,29 @@ abstract class DatabaseQuery implements QueryInterface
 	}
 
 	/**
+	 * Add a WHERE NOT IN statement to the query.
+	 *
+	 * Note that all values must be the same data type.
+	 *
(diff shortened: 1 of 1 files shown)


HLeithner:
@mbabker I would like to see this merged, as long as we don't have a better solution for such functions.

* extendWhere and it's short cuts have the same problem...

PR #161, opened 2019-05-30
Add named prepared statements to sqlsrv

HLeithner:
Pull Request for Issue #112 and tries to bring named prepared statements to the sqlsrv driver.

### Summary of Changes
* bring sqlsrv driver more in sync with mysqli driver
* add named prepared statements
* fixed wrong docblocks

### Testing Instructions
* use (named) prepared statements with mssql server

### DIFF
--- a/src/Sqlsrv/SqlsrvQuery.php
+++ b/src/Sqlsrv/SqlsrvQuery.php
@@ -12,6 +12,11 @@ use Joomla\Database\DatabaseInterface;
 use Joomla\Database\DatabaseQuery;
 use Joomla\Database\ParameterType;
 use Joomla\Database\Query\QueryElement;
+use function SQLSRV_SQLTYPE_BINARY;
+use const SQLSRV_SQLTYPE_BIT;
+use const SQLSRV_SQLTYPE_INT;
+use function SQLSRV_SQLTYPE_VARBINARY;
+use function SQLSRV_SQLTYPE_VARCHAR;
 
 /**
  * SQL Server Query Building Class.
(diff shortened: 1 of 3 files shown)


src/Sqlsrv/SqlsrvQuery.php
```diff
@@ -320,9 +361,15 @@ public function bind($key = null, &$value = null, $dataType = ParameterType::STR
 			return $this;
 		}
 
+		// Validate parameter type
+		if (!isset($this->parameterMapping[$dataType]))
+		{
+			throw new \InvalidArgumentException(sprintf('Unsupported parameter type `%s`', $dataType));
```
mbabker: We should probably add our own `Joomla\Database\Exception\InvalidParameterException extends \InvalidArgumentException {}` subclass since this check exists in basically every query class, that way that specific error can be caught in isolation.

HLeithner: I will do this after #162 is merged while updating in sqlsrv driver.

HLeithner:
@mbabker can you help me with the tests?

why the tests load the sqlsrv files?
Also why are the php 7.2 is only tested with sqlite?

https://travis-ci.org/joomla-framework/database/builds/539741133

mbabker:
> why the tests load the sqlsrv files?

The query class tests for the most part never relied on driver specific code, so those never had anything to skip them.  That's probably going to be needed for the SQL Server query class now.

> why are the php 7.2 is only tested with sqlite?

Pull requests run a much smaller subset of the builds (basically one build per DB driver and if necessary server version), compare your build with https://travis-ci.org/joomla-framework/database/builds/538952534 which has the full matrix.  The intent is if it work on lowest and highest deps (which is what the subset should mostly cover), it should work on everything in between.

Also note that the builds aren't only running the SQLite or MySQL or PostgreSQL or SQL Server tests.  That's basically stating what database backend is configured for that build (with SQLite always being there because it's baked into the PHP build and the container image) and everything that can run in that configured environment does.  This is another reason the tests need a major overhaul, or at least proper grouping.

mbabker:
Can you sync your branch, then we can see where things stand here?

HLeithner:
I can close this one because it's already included with PR #164

PR #162, opened 2019-05-30
Accept array in bind function and reduce duplicated code

HLeithner:
Add array syntax to databaseQuery::bind function

### Summary of Changes
Reduced duplicated code.
Does not include sqlsrv because of #161 a separate PR is needed

### Testing Instructions
Use bind and bindArray with arrays as key, value and datatype.


src/DatabaseQuery.php
```diff
@@ -1793,6 +1807,95 @@ public function andWhere($conditions, $glue = 'OR')
...
+				if (isset($this->bounded[$key[$i]]))
+				{
+					unset($this->bounded[$key[$i]]);
+				}
+
+				continue;
+			}
+
+			// Validate parameter type
+			if (!isset($this->parameterMapping[$localDataType]))
```
mbabker: Either this needs to be defined as a property on the main query class or you'll need to do a `isset($this->parameterMapping)` check first otherwise you can get an undefined property message.

HLeithner: thx, I missed it.

HLeithner:
#161 have to be merged before this can be merged.

HLeithner:
I close this one because it's already included in #163

HLeithner:
Mixed the PR with another one, this is still valid and now up to date.

HLeithner:
@mbabker can you explain me for what the query mock is and why it has a copy of the bind function?

mbabker:
It's a test stub, and probably has the methods it did because they didn't exist in the root query class before.

HLeithner:
So I can remove it?

mbabker:
Probably, if there's no need for it.

alikon:
i've tested with pdo postgresql .... and with this pr https://github.com/joomla/joomla-cms/pull/25043
so the new syntax is working as intented

HLeithner:
@wilsonge @mbabker anyone like to merge this?

mbabker:
I'll look later (about to head out), the CS build has some corrections that need to be made though.  https://travis-ci.org/joomla-framework/database/jobs/554133491

HLeithner:
thx

PR #163, opened 2019-06-05
[RFC] Move parameterTypeMapping from Query to Statement

HLeithner:
### Summary of Changes
For consistent and better sqlsrv support move parameter mapping from query to statement class


HLeithner:
@mbabker any feedback is welcome

PR #164, opened 2019-06-05
[RFC] Sqlsrv prepared statement based on #163

HLeithner:
### Summary of Changes
This is based on #161 and #163 for more consistent in the driver API and less overhead in prepare method.


HLeithner:
@mbabker any feedback is welcome

PR #166, opened 2019-06-20
[2.0] [WiP] MariaDB support for MySQLi and MySQL (PDO) drivers

richard67:
Pull Request for Joomla Issue [https://github.com/joomla/joomla-cms/issues/25245](https://github.com/joomla/joomla-cms/issues/25245).

### Summary of Changes

Add version check compatibility for MariaDB to MySQLi and MySQL (PDO) drivers.

### Testing Instructions

Apply the changes in this PR to the corresponding files in folder `libraries/vendor/joomla/database/src` and sub-folders on a current 4.0-dev branch of Joomla CMS.

Check new installation with MySQLi and MySQL (PDO) works as well as before on a MySQL database with version >= 5.6.

Check new installation with MySQLi and MySQL (PDO) works now on a MariaDB database with version >= 10.0.

Plus do a code review.

No.

### DIFF
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -56,12 +56,28 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	protected $utf8mb4 = false;
 
 	/**
+	 * True if the database engine is MariaDB.
+	 *
+	 * @var    boolean
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $mariadb = false;
+
+	/**
 	 * The minimum supported database version.
 	 *
 	 * @var    string
 	 * @since  1.0
 	 */
 	protected static $dbMinimum = '5.6';
+
+	/**
+	 * The minimum supported MariaDb database version.
+	 *
+	 * @var    string
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected static $dbMinMariadb = '10.2';
 
 	/**
 	 * Constructor.
@@ -139,11 +155,19 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 			parent::connect();
 		}
 
+		$serverVersion = $this->getVersion();
+
+		$this->mariadb = stripos($serverVersion, 'mariadb') !== false;
+
 		if ($this->utf8mb4)
 		{
 			// At this point we know the client supports utf8mb4.  Now we must check if the server supports utf8mb4 as well.
-			$serverVersion = $this->getVersion();
 			$this->utf8mb4 = version_compare($serverVersion, '5.5.3', '>=');
+
+			if ($this->mariadb && version_compare($server_version, '10.0.0', '<'))
+			{
+				$this->utf8mb4 = false;
+			}
 
 			if (!$this->utf8mb4)
 			{
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -425,6 +425,18 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	}
 
 	/**
+	 * Get the minimum supported database version.
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getMinimum()
+	{
+		return $this->mariadb ? static::dbMinMariadb : static::$dbMinimum;
+	}
+
+	/**
 	 * Determine whether the database engine support the UTF-8 Multibyte (utf8mb4) character encoding.
 	 *
 	 * @return  boolean  True if the database engine supports UTF-8 Multibyte.
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -424,6 +424,29 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 		return $this->setQuery('SHOW TABLES')->loadColumn();
 	}
 
+
+	/**
+	 * Get the version of the database connector.
+	 *
+	 * @return  string  The database connector version.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getVersion()
+	{
+		$this->connect();
+
+		$version = $this->getOption(\PDO::ATTR_SERVER_VERSION);
+
+		if (stripos($version, 'mariadb') !== false)
+		{
+			// MariaDB: Strip off any leading '5.5.5-', if present
+			return preg_replace('/^5\.5\.5-/', '', $version);
+		}
+
+		return $version;
+	}
+
 	/**
 	 * Get the minimum supported database version.
 	 *
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -433,7 +433,7 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	 */
 	public function getMinimum()
 	{
-		return $this->mariadb ? static::dbMinMariadb : static::$dbMinimum;
+		return $this->mariadb ? static::$dbMinMariadb : static::$dbMinimum;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/Mysqli/MysqliDriver.php
```diff
@@ -529,9 +547,27 @@ public function getVersion()
 	{
 		$this->connect();
 
+		if ($this->mariadb)
+		{
+			// MariaDB: Strip off any leading '5.5.5-', if present
+			return preg_replace('/^5\.5\.5-/', '', $this->connection->server_info);
```
mbabker: It needs to also account for if the version is suffixed with "-mariadb".  [Doing this](https://github.com/doctrine/dbal/blob/91f01420e1eaf6cf746ba45ccd3880a571946c32/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php#L98-L118) should be acceptable.

src/Pdo/PdoDriver.php
```diff
@@ -459,7 +459,15 @@ public function getVersion()
 	{
 		$this->connect();
 
-		return $this->getOption(\PDO::ATTR_SERVER_VERSION);
+		$version = $this->getOption(\PDO::ATTR_SERVER_VERSION);
+
+		if (stripos($version, 'mariadb') !== false)
```
mbabker: This really belongs in the `MysqlDriver` subclass, not the generic PDO class.

src/Mysqli/MysqliDriver.php
richard67: @mbabker Check for `mariadb` is done before when setting the `$this->mariadb` flag in the `connect()` function. Isn't that sufficient or unsafe somhow? If so, I'll change it of course.

src/Pdo/PdoDriver.php
richard67: Then I have to override the `getVersion` there, right?

src/Mysqli/MysqliDriver.php
mbabker: I'm not talking about the flag.  I'm talking about processing the version string.  Right now you're removing the "5.5.5-" prefix.  With a string like "5.5.5-10.1.29-MariaDB" that means this method is returning "10.1.29-MariaDB" right now, IMO that's not what we want.

src/Pdo/PdoDriver.php
mbabker: Yes.

src/Mysqli/MysqliDriver.php
richard67: But that's what we currently have. The version strings are like that ("10.1.29-MariaDB") for MySQL, too, because in MySQLi driver `server_info` and not `server_version` is used.

richard67: I agree it would be better to have numerical version checks but we don't have that right now, and I wanted to limit impact of this PR to the MariaDB thing.

src/Pdo/PdoDriver.php
richard67: Ok, will do so. It may take a while, I have to do something else for a few hours (family stuff).

src/Mysql/MysqlDriver.php
```diff
@@ -400,6 +424,18 @@ public function getTableList()
...
+	/**
+	 * Get the minimum supported database version.
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getMinimum()
+	{
+		return $this->mariadb ? static::dbMinMariadb : static::$dbMinimum;
```
Quy:
```suggestion
		return $this->mariadb ? static::$dbMinMariadb : static::$dbMinimum;
```

src/Mysqli/MysqliDriver.php
```diff
@@ -529,9 +547,27 @@ public function getVersion()
...
+	/**
+	 * Get the minimum supported database version.
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getMinimum()
+	{
+		return $this->mariadb ? static::dbMinMariadb : static::$dbMinimum;
```
Quy:
```suggestion
		return $this->mariadb ? static::$dbMinMariadb : static::$dbMinimum;
```

src/Mysqli/MysqliDriver.php
richard67: @mbabker Is that ok like I wrote above, or do you still want me to change that?

src/Mysql/MysqlDriver.php
```diff
@@ -400,6 +424,41 @@ public function getTableList()
 		return $this->setQuery('SHOW TABLES')->loadColumn();
 	}
 
+
```
Quy:
```suggestion
```

richard67:
Is it my fault that Travis failed? I would feel more comfortable if it was Elvis who had failed ;-)

src/Mysql/MysqlDriver.php
```diff
@@ -63,6 +71,14 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	 */
 	protected static $dbMinimum = '5.6';
 
+	/**
+	 * The minimum supported MariaDb database version.
```
Quy:
```suggestion
	 * The minimum supported MariaDB database version.
```

src/Mysqli/MysqliDriver.php
```diff
@@ -70,13 +70,29 @@ class MysqliDriver extends DatabaseDriver implements UTF8MB4SupportInterface
...
+	/**
+	 * The minimum supported MySQL database version.
 	 *
 	 * @var    string
 	 * @since  1.0
 	 */
 	protected static $dbMinimum = '5.6';
 
+	/**
+	 * The minimum supported MariaDb database version.
```
Quy:
```suggestion
	 * The minimum supported MariaDB database version.
```

src/Mysqli/MysqliDriver.php
mbabker: This is fine for now I guess.

richard67:
@mbabker @Quy Thank you very much. I am really happy that you both reviewed this PR.

HLeithner:
You asked in the other issue if we still need Utf8mb4 check.
If I understand it correctly michael increased the minimum MySQL Server Version to 5.6 (as the CMS) so all Servers we connect to will always have Utf8mb4 is thsi correct @mbabker

richard67:
@HLeithner : @mbabker meanwhile answered here to that in a code review discussion which is hidden meanwhile because as solved. As far as I remember he said it is ok now.

mbabker:
I purposefully left the utf8mb4 checks alone for now, in part because I don't really know what MariaDB versions do or do not support the character set.  But yes, eventually that can be cleaned up so that the checks that are based on the MySQL server version number are removed (as those are obsolete).  Other checks based on MariaDB compatibility (if necessary) or the client version (the C level library talking to the server, libmysql or mysqlnd typically) need to stay in place.

richard67:
The check has already been corrected in this PR. For safety I took 10.0, which as far as I remember having seen in their docs already supported utf8mb4. So the test would still work and allow connections e.g. by 3rd party components to external databases others than the one on which Joomla is installed for external databases older than minimum version required for Joomla install. I think we should even leave that, it is not a bad thing.

mbabker:
The MySQL drivers are declaring they don't support MySQL 5.5 at all, the minimum property isn't just for the CMS' use (but we don't throw a hard exception on connect if the server is below that version, IMO we shouldn't either).  So there are a couple of checks that are dead code, that's what I'm getting at (because if we know everything using the driver requires a MySQL 5.6 server, checking if the server version is 5.5.3 or newer is redundant).

richard67:
Ah, ok, I see. Well we can clean that up later.

richard67:
@mbabker I guess it needs now a PR in the 4.0-dev branch of CMS in order to update the database package, right? If so, I can do that.

mbabker:
Yup.

richard67:
Ok, am starting. Thanks for help and everything.

PR #168, opened 2019-06-21
Add isMariaDb public function to the MySQL drivers

richard67:
Pull Request for no issue. Just for future use.

### Summary of Changes

As the title sais.

### Testing Instructions

Code review.

None.

### DIFF
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -473,11 +473,11 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	/**
 	 * Determine if the database engine is MariaDB.
 	 *
-	 * @return  boolean  True if the database engine is MariaDB.
+	 * @return  boolean
 	 *
 	 * @since   __DEPLOY_VERSION__
 	 */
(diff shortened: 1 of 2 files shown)


mbabker:
These methods need to call `$this->connect()` first because this check can only be made after a connection is established (the API uses lazy connections and doesn't arbitrarily open a connection when you instantiate the driver, so `(new MysqlDriver)->isMariaDb();` with a MariaDB backend would report false before connect is called and true afterwards).

richard67:
@mbabker I see. But then the same applies for the "hasUTF8mb4Support", too, which is above the new function, right? Shall I correct both or only the new one?

mbabker:
Just the new one for now.

mbabker:
(While technically it's true utf8mb4 can only be reliably determined after the connection is made, before connection it can be assumed from the configuration given to the driver, whereas with MariaDB checks there's just no way to know anything without a connection, all this to say the has method should probably be fixed but it's not as "critically" broken as this one is without a connection)

richard67:
Thank for the explanation. Yes, that is the difference.

PR #169, opened 2019-06-25
Undefined variable

SharkyKZ:
Merge on review.


richard67:
Oh, my copy and paste mistake (blush).

src/Mysql/MysqlDriver.php
```diff
@@ -164,7 +164,7 @@ public function connect()
 			// At this point we know the client supports utf8mb4.  Now we must check if the server supports utf8mb4 as well.
 			$this->utf8mb4 = version_compare($serverVersion, '5.5.3', '>=');
 
-			if ($this->mariadb && version_compare($server_version, '10.0.0', '<'))
```
wilsonge: Isn't this useless anyhow as our `dbMinMariadb` is 10.0?

mbabker: The whole thing still needs cleaning up in general.  It's fine for now.

PR #171, opened 2019-07-01
Fix named prepared statement polyfill

HLeithner:
If you have multiple namedprepared parameters with the same substring in one section of a query the prepared statement polyfill fails.

### Summary of Changes
Only replace the first occur of a found named parameter.

### Testing Instructions
Create a query like this:
```php
$db = JFactory::getDbo();

$query = $db->getQuery(true);

$query->select('id')->from('#__content')->whereIn('id', [1,2,3,4,5,6,7,8,9,10,11]);

$db->setQuery($query);

$list = $db->loadObjectList();
```

### DIFF
--- a/Tests/DatabaseQueryTest.php
+++ b/Tests/DatabaseQueryTest.php
@@ -1818,10 +1818,13 @@ class DatabaseQueryTest extends TestCase
 	 */
 	public function testWhereIn()
 	{
-		$this->instance->whereIn('id', [1, 2, 3]);
+		$this->instance->whereIn('id', [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]);
 		$this->assertThat(
 			trim(TestHelper::getValue($this->instance, 'where')),
-			$this->equalTo('WHERE id IN (:preparedArray1,:preparedArray2,:preparedArray3)'),
+			$this->equalTo(
+				'WHERE id IN (:preparedArray1,:preparedArray2,:preparedArray3,:preparedArray4,:preparedArray5,:preparedArray6,'
+				. ':preparedArray7,:preparedArray8,:preparedArray9,:preparedArray10,:preparedArray11,:preparedArray12)'
+			),
 			'Tests rendered value.'
 		);
 	}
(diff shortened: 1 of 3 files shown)


Tests/DatabaseQueryTest.php
```diff
@@ -1818,10 +1818,13 @@ public function testWhere()
 	 */
 	public function testWhereIn()
 	{
-		$this->instance->whereIn('id', [1, 2, 3]);
+		$this->instance->whereIn('id', [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]);
```
wilsonge: This test is doing nothing - this part is being correctly generated already. please see https://github.com/joomla-framework/database/pull/172 for the correct tests

src/Mysqli/MysqliStatement.php
```diff
@@ -321,7 +326,7 @@ public function bindParam($parameter, &$variable, $dataType = ParameterType::STR
 	private function bindValues(array $values)
 	{
 		$params = [];
-		$types  = str_repeat('s', \count($values));
```
wilsonge: This is an incorrect change

src/Mysqli/MysqliStatement.php
```diff
@@ -467,7 +472,7 @@ public function execute($parameters = null)
 		{
 			$this->statement->store_result();
 
-			$this->rowBindedValues = array_fill(0, \count($this->columnNames), null);
+			$this->rowBindedValues = array_fill(0, count($this->columnNames), null);
```
wilsonge: so is this

Tests/DatabaseQueryTest.php
HLeithner: I know I wanted to test first with travis

mbabker:
Please merge the `2.0-dev` branch in here to get the tests George added.

wilsonge:
I wrote some extra local tests (which I'll PR in a bit) and I think this is working as expected. Merging and we can amend as required in the future with more tests

PR #176, opened 2019-07-08
Add groupConcat function to QueryInterface

twister65:
Pull Request for Review https://github.com/joomla/joomla-cms/pull/20095/files#r300880121

### Summary of Changes
Add groupConcat function to QueryInterface

### Testing Instructions
Replace:
https://github.com/joomla/joomla-cms/blob/54c3a76adcfe5d1060f8a6f337c6a4fdfaab294f/libraries/src/Console/ListUserCommand.php#L64-L66

By: 
`$query = $db->getQuery(true);`
`$query`
`->select($db->quoteName(['u.id', 'u.username', 'u.name', 'u.email', 'u.block']))`
`->select($query->groupConcat($db->quoteName('g.group_id')) . ' AS ' . $db->quoteName('groups'))`

And execute:
> php cli/joomla.php user:list

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -1106,21 +1106,6 @@ abstract class DatabaseQuery implements QueryInterface
 	}
 
 	/**
-	 * Aggregate function to get input values concatenated into a string, separated by delimiter
-	 *
-	 * Usage:
-	 * $query->group_concat('id', ',');
-	 *
-	 * @param   string  $column      The name of the column to be concatenated.
-	 * @param   string  $separator   The delimiter of each concatenated value
-	 *
-	 * @return  string
-	 *
-	 * @since   __DEPLOY_VERSION__
-	 */
-	abstract public function group_concat($column, $separator = ',');
-
-	/**
 	 * A conditions to the HAVING clause of the query.
 	 *
 	 * Usage:
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -1106,6 +1106,21 @@ abstract class DatabaseQuery implements QueryInterface
 	}
 
 	/**
+	 * Aggregate function to get input values concatenated into a string, separated by delimiter
+	 *
+	 * Usage:
+	 * $query->group_concat('id', ',');
+	 *
+	 * @param   string  $column      The name of the column to be concatenated.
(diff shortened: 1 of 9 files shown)


HLeithner:
Does sqlsrv has such function too?

src/DatabaseQuery.php
```diff
@@ -1105,6 +1105,21 @@ public function group($columns)
...
+	 * $query->group_concat('id', ',');
+	 *
+	 * @param   string  $column      The name of the column to be concatenated.
+	 * @param   string  $separator   The delimiter of each concatenated value
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	abstract public function group_concat($column, $separator = ',');
```
mbabker: Should be made part of the interface and the method name camel cased.

mbabker:
> Does sqlsrv has such function too?

https://database.guide/the-sql-server-equivalent-to-group_concat/

mbabker:
You also need to find an implementation for SQLite.

src/DatabaseQuery.php
twister65: Done. See https://github.com/joomla-framework/database/pull/176/commits/3157f06d48e9299892db15b9d35926a08140baa0

twister65:
Yes, it must also be implemented in sqlsrv and sqlite.
What about the sqlsrv version ?

twister65:
I have tested this PR with MySQL and PostgreSQL.
It needs more tests with SQLite (https://www.w3resource.com/sqlite/aggregate-functions-and-grouping-group_concat.php) and SqlSrv.

mbabker:
There's a test stub in `Tests/Stubs/nosqldriver.php` you'll need to add the new method to as well.  Otherwise, PR seems fine for what it is.

> It needs more tests with SQLite

Realistically, adding unit tests for the query classes would be enough, but the tests are in such rough shape I wouldn't worry about it for the moment.  As long as we're not getting a parser error we're good here.

> What about the sqlsrv version ?

Figure it out later.  If someone decides to bump past SQL Server 2012 but not all the way to SQL Server 2017 then we can figure out how to not cause this to spit errors.

f-hamel:
@twister65 Great PR. This could be very helpful in my PR, you referenced.
I've also test this sucessfully.

Issue #157, opened 2019-05-07
Port required from Joomla CMS


wilsonge:
https://github.com/joomla/joomla-cms/pull/12417

richard67:
@wilsonge For https://github.com/joomla/joomla-cms/pull/24999 see #217 . Will this then be merged up into 2.0-dev? Or shall I make a separate PR for 2.0-dev?

richard67:
@wilsonge For joomla/joomla-cms#12417 see #219 .

richard67:
Can be closed as having these 2 PRs.

PR #179, opened 2019-08-03
Add a castAs method

wilsonge:
This method replaces `castAsChar` which had fallen behind the CMS in terms of support anyhow. This uses the concept proposed by @alikon in https://github.com/joomla/joomla-cms/pull/18961#issuecomment-406776272

Includes tests & docs

Somewhat related is https://github.com/joomla-framework/database/issues/157 over which matches some of the castAsChar stuff to gain complete sync with the CMS on the `castAsChar` method

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -9,6 +9,7 @@
 namespace Joomla\Database;
 
 use Joomla\Database\Exception\QueryTypeAlreadyDefinedException;
+use Joomla\Database\Exception\UnknownTypeException;
 
 /**
  * Joomla Framework Query Building Class.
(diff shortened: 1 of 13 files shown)


mbabker:
@wilsonge Care to rebase this?

alikon:
tested ok on postgresql 11.5

alikon:
conflict apart can we have a response here ?

wilsonge:
Conflicts fixed (hopefully) - they were complicated and then I just clean forgot about this. If the tests pass I think this should be good to merge

src/DatabaseQuery.php
```diff
@@ -555,6 +556,35 @@ public function call($columns)
...
+	 */
+	public function castAs(string $type, string $value, ?string $length = null)
+	{
+		switch (strtoupper($type))
+		{
+			case 'CHAR':
+				return $value;
+
+			default:
+				throw new UnknownTypeException;
```
mbabker: Add a message to the exception please

wilsonge: Done

PR #180, opened 2019-08-05
Fixing default values from getTableColumns

Hackwar:
This PR stems from an error thrown by the tests in this PR: https://github.com/joomla/joomla-cms/pull/25761

The getTableColumns() method for Postgres returns some funky data for at least the datetime column type. The default for the given column is '1970-01-01 00:00:00' (as wrong as that is as a null value...), but the default value given from getTableColumns (and subsequently used by JTable) is "'1970-01-01 00:00:00'::timestamp without time zone". This tries to fix that. I also think that the null-code around line 228 has to be moved up like I did, otherwise it wont do anything... With this code, the PR mentioned above, passes.


wilsonge:
Can you rebase this onto 2.0-dev please now

Hackwar:
There you go.

wilsonge:
Thanks!

PR #181, opened 2019-08-10
Fix unit tests

wilsonge:
In some of our unit tests on postgres we're die'ing (which is showing as a test pass). So swap this for a `\LogicException` in the `connected` method. Finally in `execute` add a new extra try/except to catch this exception and throw the documented `\Joomla\Database\Exception\ExecutionFailureException`


Hackwar:
Together with #180 and #182, this should fix the postgres tests and fix all issues that came up with this.

PR #182, opened 2019-08-10
Removing double LIMIT for special queries

Hackwar:
At the point when this PR was created, our unittests for Postgres were failing silently in the testQuerySetWithUnionAll() test. This was due to a broken query created by the query builder.

The test uses $query->querySet(), which set the type of the query to 'querySet'. This in return resulted in an issue in the __toString() method. Based on the query type, a certain path in the switch-statement is used and 'querySet' is falling through to the default case. The default case is calling the parent builder and the parent builder processes the LIMIT at the end. Unfortunately the query builder also did the LIMIT processing at the end unconditionally. This PR fixes that. 

After this, the tests should properly pass again. While this fixes the initial error, this still does not fix the broken error handling that resulted in the tests failing silently.

### DIFF
--- a/src/Pgsql/PgsqlQuery.php
+++ b/src/Pgsql/PgsqlQuery.php
@@ -110,6 +110,8 @@ class PgsqlQuery extends PdoQuery
 					$query .= (string) $this->noWait;
 				}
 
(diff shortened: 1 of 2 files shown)


wilsonge:
Thanks!

PR #185, opened 2019-08-19
Use $table as argument for getAlterTableSql call

richard67:
Pull Request for Issue [https://github.com/joomla/joomla-cms/issues/22620](https://github.com/joomla/joomla-cms/issues/22620).

Back port of PR [https://github.com/joomla/joomla-cms/pull/22660](https://github.com/joomla/joomla-cms/pull/22660).

### Summary of Changes

In function `mergeStructure` of the importer, use `$table` and not `$xml->database->table_structure` as arugment for call to function `getAlterTableSql` for importing not only the 1st table in case of multiple tables.

### Testing Instructions

Either code review, or create some code which calls the exporter to export a few tables to XML, and some other code to call the importer to importthat (for testing possibly modified) XML, and execute that code and check if all works, and all tables are imported.

No.

### Additional Information

At a first view it seems there is no test yet for doing complete export and import to/from XML with multiple tables. I will investigate that further.

### DIFF
--- a/src/DatabaseImporter.php
+++ b/src/DatabaseImporter.php
@@ -261,7 +261,7 @@ abstract class DatabaseImporter
 			if (\in_array($tableName, $tables, true))
 			{
 				// The table already exists. Now check if there is any difference.
-				if ($queries = $this->getAlterTableSql($xml->database->table_structure))
+				if ($queries = $this->getAlterTableSql($table))
 				{
 					// Run the queries to upgrade the data structure.
 					foreach ($queries as $query)


richard67:
@mbabker I haven't tested that yet with a real test, would have to write some code for that but it might take a while until I have time for that. I did code review and checked that the called function expects that parameter in the right way. So feel free to merge on review or wait until I report here that I have tested it.

Btw. test: See section "Additional Information" in the description. Does it need to write some test? If yes: Should I do that in this PR or a new one? I would do my best but also can become weekend before I can do that.

mbabker:
TBH, I personally wouldn't invest too much time in writing new tests right now.  The entire test suite needs to be rewritten to break the dependency on a deprecated PHPUnit extension and at that point if someone actually gets around to doing that I would just tell them to write a new test suite and build it out properly instead of the haphazard and mediocre test suite that exists now (this entire package should be functional test oriented, instead we have unit tests making sure a SELECT statement gets added to the internal `QueryElement` for every `DatabaseQuery` subclass 🤦‍♂).

richard67:
@mbabker I see, well yes, sounds weird. By the way the change from this PR here should be merged up to 2.0-dev, the error exists there, too.

PR #187, opened 2019-08-21
redis driver

alikon:
### Summary of Changes
give the chance  to use redis as 1st citizen db

### Testing Instructions
simply use 
`use Joomla\Database\Redis\RedisDriver;`

added a pr to the cms repo  https://github.com/joomla/joomla-cms/pull/25957 to use this  in /..libraries/src/Cache/Storage/RedisStorage.php

### DIFF
--- a/src/Redis/RedisDriver.php
+++ b/src/Redis/RedisDriver.php
@@ -1,0 +1,216 @@
+<?php
+/**
+ * Part of the Joomla Framework Database Package
+ *
+ * @copyright  Copyright (C) 2005 - 2019 Open Source Matters, Inc. All rights reserved.
+ * @license    GNU General Public License version 2 or later; see LICENSE
+ */
+
+namespace Joomla\Database\Redis;
+
+use Joomla\CMS\Factory;
+use Psr\Log;
+use Redis;
+
+
+/**
+ * Redis Database Driver
+ *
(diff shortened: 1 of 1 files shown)


src/Redis/RedisDriver.php
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * Part of the Joomla Framework Database Package
+ *
+ * @copyright  Copyright (C) 2005 - 2019 Open Source Matters, Inc. All rights reserved.
+ * @license    GNU General Public License version 2 or later; see LICENSE
+ */
+
+namespace Joomla\Database\Redis;
+
+use Joomla\CMS\Factory;
```
SharkyKZ: Can't use CMS classes here.

mbabker:
Not quite sure what the aim is here?  This looks like it's just a factory class to create a `\Redis` instance, it's not really integrating with anything else.

mbabker:
(Also considering this package is geared toward an API for RDBMS, not sure how you would add in drivers for systems like MongoDB or Redis that integrate into the existing API)

wilsonge:
What Michael said - this package is based around SQL packages - having this in here which doesn't implement QueryInterface or DatabaseInterface makes no sense

PR #193, opened 2019-10-13
Allow the use of UNIX socket with unix: URI

andrepereiradasilva:
### Summary of Changes
- Allow to use unix socket URI (ex: `unix:/path/to/unix/socket.sock`) in PDO MySql/PDO PostgreSQL and MySqli drivers
- Allows to use unix sockets in PDO MySql
- Just this has added in this PR, the rest should work as before

### Testing Instructions

- Code review
- Use joomla latest 4.0 version
- Apply this patch to the framework database package
- Try to connect to the database with PDO Mysql/PDO PostgreSQL and MySqli DB connectors using a UNIX socket URI in the database "Host" field in global config

Examples of UNIX socket URI:
`unix:/var/run/mysqld/mysqld.sock`
`unix:/var/run/mariadb/mariadb.sock`
`unix:/var/run/postgresql/`

### Additional information

*MySQL/MariaDB*
- At least in CentOS, MySql socket path is in `/etc/my.cnf` (or `/etc/my.cnf.d/*.conf`)
- For Mysql i also added the same path in php ini config
```ini
mysqli.default_socket    = /var/run/mariadb/mariadb.sock
pdo_mysql.default_socket = /var/run/mariadb/mariadb.sock
```
*PostgreSQL*
- At least in CentOS, PostgreSQL socket path is in `/var/lib/pgsql/data/postgresql.conf` (or `/var/lib/pgsql/<version>/data/postgresql.conf`)
- In postgresql you will probably need to give permission for the user to use UNIX sockets. At least in CentOS is in `/var/lib/pgsql/data/pg_hba.conf` (or `/var/lib/pgsql/<version>/data/pg_hba.conf`)
Replace `local   all             all                                     peer` with `local   all             all                                     md5` (replace `peer` by `md5`) and restart postgresql

### DIFF
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -17,6 +17,7 @@ use Joomla\Database\Exception\PrepareStatementFailureException;
 use Joomla\Database\Exception\UnsupportedAdapterException;
 use Joomla\Database\StatementInterface;
 use Joomla\Database\UTF8MB4SupportInterface;
+
 /**
  * MySQLi Database Driver
  *
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -7,7 +7,6 @@
  */
 
 namespace Joomla\Database\Mysqli;
-
 use Joomla\Database\DatabaseDriver;
 use Joomla\Database\DatabaseEvents;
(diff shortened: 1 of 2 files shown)


wilsonge:
Looks good to me - thanks and sorry for holding so long on this!

PR #195, opened 2019-10-13
Refactor argument types in the database factory, importer, and exporter

mbabker:
### Summary of Changes

The database factory, exporter, and importer are not supportive of the `DatabaseInterface` at the moment, therefore these objects are refactored to support them.  This results in B/C breaking signature changes as the typehinted driver is changed (`Joomla\Database\DatabaseDriver` to `Joomla\Database\DatabaseInterface`).  Since I changed basically all the factory signatures, I went ahead and applied full typehinting to the class' API as well.

I've also removed the catch when instantiating a database driver through the factory.  This looks to be a relic of a time past when the drivers did not support a lazy connection and opened the connection in the constructors, right now in the provided adapters nothing does anything that would throw a `RuntimeException` so this is basically a bit of dead code.

### Testing Instructions

- Factory still builds objects correctly when given proper parameters
- Importer and exporter can still receive database drivers

- Note changed signatures in the factory

### DIFF
--- a/src/DatabaseExporter.php
+++ b/src/DatabaseExporter.php
@@ -204,15 +204,15 @@ abstract class DatabaseExporter
 	}
 
 	/**
-	 * Sets the database connector to use for exporting structure and/or data from MySQL.
+	 * Sets the database connector to use for importing structure and/or data.
 	 *
-	 * @param   DatabaseDriver  $db  The database connector.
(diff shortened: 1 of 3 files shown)


mbabker:
Paging @wilsonge

wilsonge:
LGTM! Thanks :)

PR #197, opened 2019-10-17
[2.0] Enforce the query type instead of only documenting a suggestion

mbabker:
### Summary of Changes

Right now there are several doc blocks indicating that changing a query type should not be done (i.e. a `$query->select('foo')->delete('bar')` type chain, but this is only a suggestion and not enforced.  This PR proposes to change this from a suggestion to an enforced behavior by throwing an exception if trying to mutate the type.  Note this doesn't break if you clear a property that sets the type before calling a method that would reset the type (i.e. a `$query->select('foo')->clear('select')->delete('bar')` chain would still be valid).

On a side note, the type inconsistency of `DatabaseQuery::$type` is a pain in the backside...

### Testing Instructions

Code review, in downstream applications queries should be reviewed for potential type reset issues.

Yep, already noted in doc blocks and upgrade guide.

### DIFF
--- a/docs/v1-to-v2-update.md
+++ b/docs/v1-to-v2-update.md
@@ -108,3 +108,9 @@ The first argument, `$table`, stops accepting the array. Only `DatabaseQuery` ob
 - Argument `$subQueryAlias` has been removed.
 
 Instead of `$query->from($subquery, 'alias')` use `$query->from($subquery->alias('alias'))`.
+
+#### Query type immutability enforced
+
+Several methods which defined a query type (i.e. INSERT, SELECT, UPDATE) contained doc blocks indicating the query type should not be changed, but
+this was not enforced in the code. As of 2.0, an exception will be thrown if trying to change a query type. If intending to change the query type,
+either call `DatabaseQuery::clear()` (optionally only clearing the type clause) or a new query object should be created instead.
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -8,6 +8,8 @@
 
 namespace Joomla\Database;
 
+use Joomla\Database\Exception\QueryTypeAlreadyDefinedException;
+
 /**
  * Joomla Framework Query Building Class.
  *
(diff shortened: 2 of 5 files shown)


mbabker:
Ping @wilsonge

docs/v1-to-v2-update.md
```diff
@@ -108,3 +108,9 @@ The first argument, `$table`, stops accepting the array. Only `DatabaseQuery` ob
 - Argument `$subQueryAlias` has been removed.
 
 Instead of `$query->from($subquery, 'alias')` use `$query->from($subquery->alias('alias'))`.
+
+#### Query type immutability enforced
+
+Several methods which defined a query type (i.e. INSERT, SELECT, UPDATE) contained doc blocks indicating the query type should not be changed, but
+this was not enforced in the code. As of 2.0, an exception will be thrown if trying to change a query type. If intending to change the query type,
+a new query object should be created instead.
```
wilsonge: or you should use the `clear` method

src/DatabaseQuery.php
```diff
@@ -1221,9 +1246,20 @@ public function having($conditions, $glue = 'AND')
...
 	 * @since   1.0
+	 * @throws  QueryTypeAlreadyDefinedException if the query type has already been defined
 	 */
 	public function insert($table, $incrementField = false)
 	{
+		if ($this->type !== null && $this->type !== '' && $this->type !== 'insert')
+		{
+			throw new QueryTypeAlreadyDefinedException(
+				\sprintf(
+					'Cannot set the query type to "insert" as the query type is already set to "%s", a new query should be created instead.',
```
HLeithner: Is there a reason you don't use the magic constant __METHOD__ ?
I know not really needed but would make the parts more equal and maybe easier to maintain (because of copy paste)

mbabker: The constant uses the FQCN.

HLeithner: ok that make no sense thx.

docs/v1-to-v2-update.md
mbabker: Changed docs and exception message.

PR #198, opened 2019-10-18
Option to output the database query without pre tags

roland-d:
### Summary of Changes
When you are developing and debugging with XDebug and want to get the query, you can do a `$query->dump()` but this includes the pre tags. You need to copy paste the query, remove the tags and then you can execute the query. With this change you can easily do `$query->dump(false)` to get the query without pre tags and just copy paste and execute.

### Testing Instructions
1. Apply patch
2. Add `$query->dump(false)` where you have a query
3. The output should not contain the pre tags.

Document that you can pass false to get rid of the pre tags

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -731,13 +731,32 @@ abstract class DatabaseQuery
 	 * Usage:
 	 * echo $query->dump();
 	 *
+	 * To get only the query without <pre> tags:
+	 * echo $query->dump(false);
+	 *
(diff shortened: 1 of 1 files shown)


mbabker:
Can we not add functionality to this deprecated function?  Like I pointed out in the PR to deprecate it, this really should have been removed from the API years ago, I'd rather not add new functionality that might encourage someone to use it.

(Also, there's no guarantee you will always have a database executable query here if you're using bound parameters)

Issue #199, opened 2019-10-23
MySQLi can't load custom class objects if they contain protected properties


SharkyKZ:
### Steps to reproduce the issue

Do `$db->loadObjectList('key', SomeClass::class)` when `SomeClass` contains protected properties.

### Expected result

Works.

### Actual result

`Trying to access protected property` error.

### Additional comments

I'm not sure whether this is expected or not because this works fine on PDO drivers.

mbabker:
It’s intended.  The object actually gets manually instantiated and seeded by the MysqliStatement class, whereas with PDO the extension handles it all (I’ve never dig in to the C code but I imagine there is some form of reflection involved to be able to direct set non public properties).  It has to be this way otherwise we lose libmysql compatibility and force mysqlnd only.

mbabker:
We lose libmysql if we try to use the `ext/mysqli` API in the "right" way because it doesn't support certain functions (namely `mysqli_stmt_get_result()`, which means a lot of that logic has to be emulated in userland code).  Look at the master branch of this repo, that does use the API the "right" way and we have checks in it that block libmysql support as a result.

(Note this is the same approach taken by Doctrine for much the same reason, this is why you're generally better off not passing a custom class into these database API functions and using some form of ORM or data mapping layer to convert the database result set to your own class object)

mbabker:
Actually, if I would have done things right when I originally implemented things, `$db->loadObjectList('key', SomeClass::class)` should end up triggering a code path that throws an Exception.  `DatabaseDriver::fetchObject()` should be setting the fetch mode to `FetchMode::CUSTOM_OBJECT` if the class isn't `stdClass`, which isn't a mode that the MySQLi driver supports and will throw an Exception.  Sure, the API technically supports loading in a custom class, but it's limited in nature (class must have public properties or a magic setter) and not feature compatible with what `ext/mysqli` or `ext/pdo` or `ext/sqlsrv` do under the hood so the right action is to trigger the error.

SharkyKZ:
https://github.com/joomla-framework/database/blob/abbfe675d9102c0ae1ec63d746783d3ac9f06dfc/src/Mysqli/MysqliStatement.php#L557-L560
What I meant is, can't we set properties here using Reflection? Or would that break libmysql? Or just be a completely wrong approach?

>$db->loadObjectList('key', SomeClass::class) should end up triggering a code path that throws an Exception.

[...]

PR #200, opened 2019-10-26
Set the correct fetch mode for non-PDO drivers

mbabker:
Pull Request for Issue #199

### Summary of Changes

We have a fetch mode for specifying a custom PHP class should be used instead of a plain `stdClass`, the `FetchMode::CUSTOM_OBJECT` (mapped to `PDO::FETCH_CLASS` for the PDO drivers).  Because of the way MySQLi is structured, it isn't actually possible to fully support custom class objects in the same way that the PHP extensions would handle business, and use of a custom class should trigger an Exception as a result.  However, the code isn't switching over to this fetch mode which means the Exception isn't being reached.  So, this PR adds a check to force the statement into the correct mode.

### DIFF
--- a/src/Mysqli/MysqliStatement.php
+++ b/src/Mysqli/MysqliStatement.php
@@ -628,7 +628,9 @@ class MysqliStatement implements StatementInterface
 	{
 		$this->defaultObjectClass = $className;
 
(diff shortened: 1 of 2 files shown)


mbabker:
Indirectly superseded by #201

PR #201, opened 2019-10-26
PdoStatement and StatementInterface changes

mbabker:
### Summary of Changes

The original implementation of the `StatementInterface` resulted in a signature set matching up with `PDOStatement` since the PDO driver's statement class was a subclass of the PDO extension.  This PR refactors the `StatementInterface` structure so that the PDO driver's statement is a decorator around the native `PDOStatement` and makes some adjustments to the `StatementInterface` as a result.

Interface changes include:

- Removing `fetchObject()`, for the non-PDO drivers this ended up as a shortcut to the `fetch()` method with the fetch mode set by default, and in the case of PDO internally it sets the fetch mode to `PDO::FETCH_CLASS` (the mapping for our `FetchMode::CUSTOM_OBJECT`).  Instead, the driver should pass forward the appropriate fetch mode before loading any rows out of the result set.
- `bindParam()` default value for `$dataType` uses the `Joomla\Database\ParameterType` class instead of `PDO`
- Added `setFetchMode()`, it's a proxy to `PDOStatement::setFetchMode()` or internally configures the non-PDO statements in the necessary way
- Typehinted API

### Testing Instructions

Things should just keep working

### DIFF
--- a/Tests/DatabaseIteratorTest.php
+++ b/Tests/DatabaseIteratorTest.php
@@ -7,6 +7,7 @@
 namespace Joomla\Database\Tests;
 
 use Joomla\Database\DatabaseIterator;
+use Joomla\Database\FetchMode;
 use Joomla\Database\StatementInterface;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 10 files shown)


SharkyKZ:
>Things should just keep working

Except for MySQLi, you mean 😄

SharkyKZ:
Works fine on MySQL PDO and PostgreSQL for me.

PR #203, opened 2019-10-29
Support binding null values

SharkyKZ:
Pull Request for Issue #202

### Summary of Changes

Adds support for binding null values.

### Testing Instructions

See #202.

Probably.

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -726,6 +726,11 @@ abstract class DatabaseQuery implements QueryInterface
 
 				break;
 
+			case 'bounded':
+				$this->bounded = [];
+
+				break;
+
 			default:
 				$this->type               = null;
 				$this->alias              = null;
@@ -1930,8 +1935,7 @@ abstract class DatabaseQuery implements QueryInterface
 	}
 
 	/**
-	 * Method to add a variable to an internal array that will be bound to a prepared SQL statement before query execution. Also
-	 * removes a variable that has been bounded from the internal bounded array when the passed in value is null.
+	 * Method to add a variable to an internal array that will be bound to a prepared SQL statement before query execution.
 	 *
 	 * @param   array|string|integer  $key            The key that will be used in your SQL query to reference the value. Usually of
 	 *                                                the form ':key', but can also be an integer.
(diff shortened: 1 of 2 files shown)


src/DatabaseQuery.php
```diff
@@ -2022,6 +2012,38 @@ public function bind($key = null, &$value = null, $dataType = ParameterType::STR
...
+	/**
+	 * Method to unbind a bound variable.
+	 *
+	 * @param   array|string|integer  $key  The key or array of keys to unbind.
+	 *
+	 * @return  $this
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function unbind($key): self
```
SharkyKZ: Should this be added to the interface?

src/DatabaseQuery.php
```diff
@@ -1947,14 +1951,11 @@ public function andWhere($conditions, $glue = 'OR')
...
+	public function bind($key, &$value, $dataType = ParameterType::STRING, $length = 0, $driverOptions = [])
 	{
-		// Case 1: Empty Key (reset $bounded array)
-		if (empty($key))
+		if (!$key)
 		{
-			$this->bounded = [];
-
-			return $this;
+			throw new \InvalidArgumentException('A key is required');
```
mbabker: Add a `@throws` annotation to the doc block.

src/DatabaseQuery.php
mbabker: Yes.

src/DatabaseQuery.php
```diff
@@ -2022,6 +2012,38 @@ public function bind($key = null, &$value = null, $dataType = ParameterType::STR
...
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function unbind($key): self
+	{
+		if (\is_array($key))
+		{
+			foreach ($key as $k)
+			{
+				if (\array_key_exists($this->bounded[$k]))
```
mbabker: You can skip the exists check, `unset()` doesn't cause any warnings if trying to unset a non-existing key.

src/DatabaseQuery.php
SharkyKZ: Added.

PR #206, opened 2019-11-17
Replace prefix only if query monitor is set

SharkyKZ:
### Summary of Changes
Removes unnecessary call when query monitor is not set.

### Testing Instructions
Everything should work like before.

No.


alikon:
make sense to me

PR #208, opened 2019-11-17
Downgrade chocolatey if required

wilsonge:
See https://github.com/chocolatey/choco/issues/1843


wilsonge:
@mbabker this passes now - should be good to merge into here and 2.0-dev

PR #209, opened 2020-01-05
Add a Zip option to the import command

twister65:
Pull Request for Comment https://github.com/joomla/joomla-cms/pull/26248#issuecomment-570782612

### Summary of Changes

- Add a --zip option to the import command that specifies the name of the zip file in addition to the folder path.

- The name of the zip export file is now unique.

### Testing Instructions

- Go to the cli folder of your website.

- Export all tables and data (in zip format) to a specified folder:
`php joomla.php database:export --all --zip --folder <folder_path>`

- Import all from a folder and a zip file:
`php joomla.php database:import --all --folder <folder_path> --zip <zip_filename>`

or

- Import a table from a folder and a zip file:
`php joomla.php database:import --table <table_name> --folder <folder_path> --zip <zip_filename>`


src/Command/ExportCommand.php
```diff
@@ -98,11 +98,9 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
 			return 1;
 		}
 
-		$date       = getdate();
-		$dateFormat = sprintf("%s-%02s-%02s", $date['year'], $date['mon'], $date['mday']);
-
-		$zipFile = $folderPath . '/data_exported_' . $dateFormat . '.zip';
-
+		$siteName = trim(strtolower($this->getApplication()->get('sitename')));
```
mbabker: You need a sensible default for the config param, it is not required that a Framework application have a “sitename” parameter in the application configuration; this is a CMS-ism.

src/Command/ImportCommand.php
```diff
@@ -55,6 +58,40 @@ public function __construct(DatabaseDriver $db)
...
+	 * @param   string  $archive  A zip archive to analyze
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  \RuntimeException
+	 */
+	private function checkZipFile($archive)
+	{
+		$zip = zip_open($archive);
```
mbabker: There should be a check here that PHP's `ext/zip` is installed and enabled otherwise this will fail out.

src/Command/ImportCommand.php
```diff
@@ -98,6 +136,49 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
...
+				$this->checkZipFile($zipPath);
+			}
+			catch (\RuntimeException $e)
+			{
+				$symfonyStyle->error($e->getMessage());
+
+				return 1;
+			}
+
+			$folderPath .= File::stripExt($zipFile);
```
mbabker: The `joomla/archive` and `joomla/filesystem` packages are dev dependencies and therefore can't be relied on to always be installed.  There should be `class_exists()` checks with appropriate error messages in case these packages are missing.

src/Command/ImportCommand.php
```diff
@@ -98,6 +136,49 @@ protected function doExecute(InputInterface $input, OutputInterface $output): in
...
+				$symfonyStyle->error($e->getMessage());
+
+				return 1;
+			}
+
+			$zipArchive = (new Archive)->getAdapter('zip');
+
+			try
+			{
+				$zipArchive->extract($zipPath, $folderPath);
```
mbabker: The right move here is actually `(new Archive)->extract($zipPath, $folderPath);`.  Generally, there aren't many reasons you need to use `Archive::getAdapter()` manually.

src/Command/ImportCommand.php
```diff
@@ -179,6 +265,7 @@ protected function configure(): void
 	{
 		$this->setDescription('Import the database');
 		$this->addOption('folder', null, InputOption::VALUE_OPTIONAL, 'Import from folder path', '.');
+		$this->addOption('zip', null, InputOption::VALUE_REQUIRED, 'Import ZIP file');
```
mbabker: This option should have a better description.  From the code it seems this is the name of a ZIP file to be imported?

PR #216, opened 2020-09-11
Moving deprecation-contracts to require

Hackwar:
When having this dependency in the require-dev section, it is not installed when the package is used as a dependency and thus the calls to trigger_deprecation() fail with an error. That breaks the tests of the authentication package.


HLeithner:
thx

Issue #222, opened 2021-01-08
CI configuration unexpectedly leaks into testing environment


mbabker:
Commits such as https://github.com/joomla-framework/database/commit/073aff4cde74bdd8f2db65c520610239d0a9ed43 have leaked environment variables used to configure the various CI platforms into the test suite.  This essentially changes the testing workflow so that it will only run correctly on a properly configured CI platform, different from the user expectation that the tests should be executable in any environment with the correct settings.

If it is intended to require this type of configuration when executing the tests, these requirements should follow existing conventions for defining constants or env vars and clearly documented for all contributors to use.  Otherwise, not documenting this mandatory change will result in unexpected behaviors if the tests are run outside of the configured CI environment.

nibra:
Thank you @mbabker for pointing this out. I was too focused on getting the build to work at all, so I completely overlooked this aspect.

PR #223, opened 2021-01-09
Fix - Don't leak CI configuration into tests

nibra:
Pull Request for Issue #222

### Summary of Changes

Use DatabaseDriver::getVersion instead of environment variable

### Testing Instructions

Code review

No


Llewellynvdm:
Do we not also need to do the same with the [above array](https://github.com/nibra/database/blob/2c6ca3e12734819d0742b4a72825ea1b93006f68/Tests/DriverMysqlTest.php#L300) `'id' => 'int unsigned'` ?

nibra:
@Llewellynvdm No, tests are working in all versions

PR #234, opened 2021-03-21
s/databse/database s/charactersets/character sets

PhilETaylor:
s/databse/database s/charactersets/character sets


nibra:
Thank you!

PR #240, opened 2021-03-22
[2.0] Dereference parameters in DebugMonitor

SharkyKZ:
### Summary of Changes

Dereferences bound parameters in `DebugMonitor` to prevent reporting wrong value when reusing the same query object.

### Testing Instructions

Set up  a connection with debug monitor.
Run the same query with different bound parameters.

Check that `DebugMonitor::getBoundParams()` returns correct values.

No.

### DIFF
--- a/src/Monitor/DebugMonitor.php
+++ b/src/Monitor/DebugMonitor.php
@@ -71,7 +71,7 @@ final class DebugMonitor implements QueryMonitorInterface
 	public function startQuery(string $sql, ?array $boundParams = null): void
 	{
 		$this->logs[]        = $sql;
-		$this->boundParams[] = $boundParams;
+		$this->boundParams[] = unserialize(serialize($boundParams));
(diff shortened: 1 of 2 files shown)


src/Monitor/DebugMonitor.php
```diff
@@ -71,7 +71,7 @@ final class DebugMonitor implements QueryMonitorInterface
 	public function startQuery(string $sql, ?array $boundParams = null): void
 	{
 		$this->logs[]        = $sql;
-		$this->boundParams[] = $boundParams;
+		$this->boundParams[] = unserialize(serialize($boundParams));
```
nibra, line 74: Please add a comment like
```php
// Dereference bound parameters to prevent reporting wrong value when reusing the same query object.
```
The code does not explain itself sufficently in this case.

nibra (changes requested):

src/Monitor/DebugMonitor.php
SharkyKZ: Added.

nibra:
Thank you!

Issue #241, opened 2021-03-24
The MySQL version check for integer display widths modifier support is not accurate enough


richard67:
### Steps to reproduce the issue

The following and possibly more changes might not be sufficient regarding the version check:

https://github.com/joomla-framework/database/commit/073aff4cde74bdd8f2db65c520610239d0a9ed43#diff-3e927448d9eaafa620c1a995e71ad8d5ab1cae1e56ef4b6a2bd78d68a430ae56R306

https://github.com/joomla-framework/database/commit/ab96e2a196a64cb99254e52a835a7d06c2f8d26f#diff-45e47bfc8339fb2db89de46a41ff0b70379e2b570c44ca9685899d898deb099cR317

MySQL who obviously don't use semver (at least not for 8.0.x) have deprecated the display width modifier beginning with version 8.0.17, see https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html .

I remember from past when I had discovered that after my MSQL had been updated so I had to make PR https://github.com/joomla/joomla-cms/pull/28501 , the display width modifier was part of the type on the 8.0.x version I had before that update.

That means the version check here should be done against version 8.0.17, and that can't be done with a floating point type cast and comparison, it needs to use something like VERSION_COMPARE for that.

Furthermore, MariaDB still support the display width modifier and so behaves like a MySQL < 8.0.17 regarding this. The version check should check this, too.

=> Ping @nibra .

richard67:
I could make a PR to fix it, but I'd use VERSION_COMPARE for that and not cast to float, but I don't know if the use of VERSION_COMPARE is maybe not desired here.

nibra:
VERSION_COMPARE is the better choice, no specific reason against it. The current implementation was chosen for brevity only. So please go ahead!

richard67:
ok, will work on it.

richard67:
I will maybe make a PR with more changes, e.g. removal of display widths from SQL scripts and statements like I do it with my 3 open PR's for the CMS. I have here a MySQL 5.7, a MySQL 8.0.20 and 2 versions of MariaDB for testing, and I want to achieve that we can run the database unit tests on all these.

If things in the 2.0-dev branch here look very different, I might make a separate PR for that branch.

richard67:
Or maybe I keep things separate and make separate PR's for this issue here and the other changes.

richard67:
Closing as having a pull request.

PR #242, opened 2021-03-24
Fix version check for display width modifier support in unit tests for MySQL

richard67:
Pull Request for Issue #241 .

### Summary of Changes

This pull request (PR) changes the 2 checks for display width modifier support in unit tests for the MySQL drivers so the check is correctly done if it's a MariaDB or a MySQL version lower than 8.0.17, which still use the display width modifier for integer data types, or if it's a MySQL 8.0.17 or later, on which the display width modifier is deprecated and not part of the column's type specifier returned e.g. by a "SHOW COLUMNS" and where it might not be supported in future.

It makes it possible to run the unit tests on MariaDB, too.

See https://github.com/joomla/joomla-cms/pull/32606 for details on the display width issue.

### Testing Instructions

Check if unit tests still pass.

None.

### Additional information

I might do a follow-up PR later to remove the usage of the display width modifier from SQL statements in scripts and at other places.

### DIFF
--- a/Tests/DriverMysqlTest.php
+++ b/Tests/DriverMysqlTest.php
@@ -311,10 +311,12 @@ class DriverMysqlTest extends DatabaseMysqlCase
 		);
 
 		/* not only type field */
+		$version = self::$driver->getVersion();
+
 		$id = new \stdClass;
 		$id->Default    = null;
 		$id->Field      = 'id';
-		$id->Type       = (float)self::$driver->getVersion() < 8.0 ? 'int(10) unsigned' : 'int unsigned';
+		$id->Type       = stripos($version, 'mariadb') || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
 		$id->Null       = 'NO';
 		$id->Key        = 'PRI';
 		$id->Collation  = null;
--- a/Tests/DriverMysqlTest.php
+++ b/Tests/DriverMysqlTest.php
@@ -316,7 +316,7 @@ class DriverMysqlTest extends DatabaseMysqlCase
 		$id = new \stdClass;
 		$id->Default    = null;
 		$id->Field      = 'id';
-		$id->Type       = stripos($version, 'mariadb') || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
+		$id->Type       = stripos($version, 'mariadb') !== false || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
 		$id->Null       = 'NO';
 		$id->Key        = 'PRI';
 		$id->Collation  = null;
--- a/Tests/DriverMysqliTest.php
+++ b/Tests/DriverMysqliTest.php
@@ -305,7 +305,7 @@ class DriverMysqliTest extends DatabaseMysqliCase
 		$id = new \stdClass;
 		$id->Default    = null;
 		$id->Field      = 'id';
-		$id->Type       = stripos($version, 'mariadb') || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
+		$id->Type       = stripos($version, 'mariadb') !== false || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
 		$id->Null       = 'NO';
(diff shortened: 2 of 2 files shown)


richard67:
@nibra Does it also need a PR for the 2.0-dev branch? Or is this something which will be merged up to there from master?

richard67:
> @nibra Does it also need a PR for the 2.0-dev branch? Or is this something which will be merged up to there from master?

Update: I just see tests have been changed there, so I have to make a separate PR.

richard67:
For 2.0-dev see #243 .

Tests/DriverMysqlTest.php
```diff
@@ -311,10 +311,12 @@ public function testGetTableColumns()
 		);
 
 		/* not only type field */
+		$version = self::$driver->getVersion();
+
 		$id = new \stdClass;
 		$id->Default    = null;
 		$id->Field      = 'id';
-		$id->Type       = (float)self::$driver->getVersion() < 8.0 ? 'int(10) unsigned' : 'int unsigned';
+		$id->Type       = stripos($version, 'mariadb') || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
```
nibra, line 319: `stripos()` needs to be tested against `false`, because 0 means that the string starts with 'mariadb'.
Doesn't it make more sense to test for MySQL, as *that* changes behaviour?

Tests/DriverMysqliTest.php
```diff
@@ -300,10 +300,12 @@ public function testGetTableColumns()
 		);
 
 		/* not only type field */
+		$version = self::$driver->getVersion();
+
 		$id = new \stdClass;
 		$id->Default    = null;
 		$id->Field      = 'id';
-		$id->Type       = (float) self::$driver->getVersion() < 8.0 ? 'int(10) unsigned' : 'int unsigned';
+		$id->Type       = stripos($version, 'mariadb') || version_compare($version, '8.0.17') < 0 ? 'int(10) unsigned' : 'int unsigned';
```
nibra, line 308: `stripos()` needs to be tested against `false`, because 0 means that the string starts with 'mariadb'.
Doesn't it make more sense to test for MySQL, as *that* changes behaviour?

nibra (changes requested):

Tests/DriverMysqliTest.php
richard67: > Doesn't it make more sense to test for MySQL, as that changes behaviour?

What do yo mean with that? Change (invert) the logic?

nibra: Yes, `stripos($version, 'mysql') && version_compare($version, '8.0.17') >= 0 ? 'int unsigned' : 'int(10) unsigned'`

richard67: This would be wrong, I think you mean:

`stripos($version, 'mysql') === false && version_compare($version, '8.0.17') >= 0 ? 'int unsigned' : 'int(10) unsigned'`

What means "does it make more sense"? Both make same sense, my current implementation here is even better because in case of mariadb has to check only one term. And I think it's better readable, too.

And in the 2.0-dev branch you have already merged that with the same logic, only difference is we could use the IsMariaDB (or so) function.

richard67: After this one will be merged, there will be one thing left to do on both branches: Fix the XML import so it strips off the display with on MySQL 8.0.17 or later. There I might use indeed inverse logic because it would make later checks not to be inverted. But that really depends on the context, what's better at that place.

nibra: `stripos($version, 'mysql') !== false` says that `$version` contains 'mysql'.
`version_compare($version, '8.0.17') >= 0` says the version is 8.0.17+.
That would change behavoiur for *MySQL 8.0.17+* and not for *any other than MariaDB with a version >= 8.0.17*, which makes the intention much more obvious.

richard67: Ah i didn't get the change with the stipos argument. Anyway, you have merged the same logic into 2.0-dev, so I see no reason to change it here.

richard67: See here for the 2.0-dev logic: https://github.com/joomla-framework/database/blob/2.0-dev/Tests/Mysql/MysqlDriverTest.php#L94

nibra:
Thank you!

PR #247, opened 2021-04-03
Update DatabaseQuery.php

joomdonation:
Pull Request for Issue #

### Summary of Changes
Made change so that bindArray method accept none-zero-index array. It can be used to fix the issue https://github.com/joomla/joomla-cms/issues/32981

### Testing Instructions
I don't know. I hope test pass is enough?

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -2098,6 +2098,9 @@ abstract class DatabaseQuery implements QueryInterface
 		{
 			$parameterNames[] = ':preparedArray' . (++$this->preparedIndex);
 		}
+		
+		// In the bind method, we expect $values in a zero-index array, so array_values is needed here
+		$values = array_values($values);
 
 		$this->bind($parameterNames, $values, $dataType);
 


richard67:
More details see https://github.com/joomla/joomla-cms/issues/32981#issuecomment-812891426 . I think a valid test here would be to update a 4.0-dev CMS to the current 20-dev branch of this framework package here and then apply the change of this PR and check that it fixes the issue.

For this, one can fetch the latest 20-dev of the database package to 4.0-dev of the CMS by changing `~2.0@beta` to `~2.0@dev` here

https://github.com/joomla/joomla-cms/blob/4.0-dev/composer.json#L43

and then running `composer update joomla/database`

and then applying the changes from this PR here manually.

joomdonation:
For someone with existing Joomla 4 installation, I think you just need to replace the existing file libraries/vendor/joomla/database/src/DatabaseQuery.php with this file https://github.com/joomla-framework/database/blob/ebcde429f8af69a99432aaf56b5ad005ac114bd0/src/DatabaseQuery.php and test to see if it fixes the CMS issue

PhilETaylor:
This looks like a PHP 8 issue. Please report your PHP version.

HLeithner:
please read https://github.com/joomla-framework/database/pull/211#issuecomment-594139754 it shouldn't be our job to fix invalid parameters. The user should supply a zero indexed array and we should guess how to fix the input.

joomdonation:
I think user should not have to worry about zero-indexed array while calling this bindArray method. Image that we start with a zero-indexed array, then perform some validation and unset some certain value from that array and then call **bindArray** method, we would get error because now the array is not sequence zero-indexed anymore (but we should not). The issue from our CMS is a perfect example for that and now we have to use workaround https://github.com/joomla/joomla-cms/pull/32990

If you don't look at this issue, do you know that you have to use sequence  zero-indexed array? I don't think so. Accepting this change will help prevent logical error like that (like the issue in CMS which we are having). Please think about it and make it easier to use our API. @nibra @wilsonge Please give some thought here.

nibra (approved):

nibra:
@joomdonation Agree

HLeithner:
Once again this could lead to an unexpected return value. See https://www.php.net/manual/de/function.array-values.php#114941

Better throw an exception then doing magic and try to fix developer invalid input

joomdonation:
> Once again this could lead to an unexpected return value. See https://www.php.net/manual/de/function.array-values.php#114941
> 
> Better throw an exception then doing magic and try to fix developer invalid input

For that special case, it is already broken, so the change here does not make it worse. If you still don't want to accept the change here, we at least should call **array_values** in whereIn v and whereNotIn https://github.com/joomla-framework/database/blob/2.0-dev/src/DatabaseQuery.php#L1903 methods because for these methods, the ordering of data in array is not important.

richard67:
> Better throw an exception then doing magic and try to fix developer invalid input

@HLeithner I guess you meant: "Better throw an exception than doing magic and try to fix developer invalid input".
Use "then" when talking about sequences in time, "first I will eat and then I will go sleep", and use "than" in comparisons, "better this than that". This small difference by 1 letter completely changes the meaning of your sentence.

@joomdonation Regarding where to use the array_values: I agree with Harald that it should not be done here. It is a kind of masking of wrong input. It could maybe make sense to fix it in the whereIn and whereNotIn, where ordering doesn't matter, but if this is done, then someone has to pick up the responsibility to clean up CMS code by the then unnecessary array_value calls.

About the cause of that problem: I have seen many places in CMS code where we use keys and values for foreach loops `foreach (array_expression as $key => $value)`, but inside the loop we never use the key, so we could have used `foreach (array_expression as $value)`, maybe just because these pieces of code where copied from somewhere else. Maybe we should clean this up in CMS code so we don't need to use array_values more often than really necessary.

PhilETaylor:
If only we had some fully automated testing framework that had some kind of code coverage that covered a range on inputs and could report the result...

It could really help prove/disprove facts and make life so much easier for developers and make conversations like this unneeded

#sarcasm

HLeithner:
> @HLeithner I guess you meant: "Better throw an exception th**a**n doing magic and try to fix developer invalid input".
> Use "then" when talking about sequences in time, "first I will eat and then I will go sleep", and use "than" in comparisons, "better this than that". This small difference by 1 letter completely changes the meaning of your sentence.

right, fixed it...

HLeithner:
On the other side I don't have much against such "magic" but it should be good documented (at least in the parameters doc block) and we should have tests for it.

for the WhereIn functions you are right at this functions it doesn't care but if we use it there we should be consistent and also allow it here...

sakiss:
@joomdonation As mentioned in my [PR](https://github.com/joomla-framework/database/pull/211), the `$values` is passed by reference. Possibly would be better to just get the values without changing the initial variable, which could be used elsewhere  thereafter.

nibra:
PR#249 solves this keeping the references.
There are no documented requirements for the indexes of the value array, so ignoring them is fine. However, array_values destroys the reference, so it is not an option. PR#249 avoids that problem.

PR #258, opened 2021-12-24
PHP 8 Deprecation Handling

PhilETaylor:
Pull Request for Issue #257

Screenshot 2021-12-24 at 17 52 52

```
40s
255	  <br />
40s
256	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>273</b><br />
40s
257	  <br />
40s
258	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>279</b><br />
40s
259	  <br />
40s
260	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>273</b><br />
40s
261	  <br />
40s
262	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>279</b><br />
40s
263	  <br />
40s
264	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>273</b><br />
40s
265	  <br />
40s
266	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>279</b><br />
40s
267	  <br />
40s
268	  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgres/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>273</b><br />
40s
269	  <br />
40s
[...]

PR #259, opened 2022-01-09
Only check types of postgres default values if set

wilsonge:
The field default can be null. If that is the case then the preg_match gives a PHP deprecation warning in PHP 8.1. 

This can be seen at https://ci.joomla.org/joomla/joomla-cms/50349/1/31 phpnext-api-postgres test run

e.g.
```
  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgresphpnext/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>273</b><br />
  <br />
  <b>Deprecated</b>:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>/tests/www/postgresphpnext/libraries/vendor/joomla/database/src/Pgsql/PgsqlDriver.php</b> on line <b>279</b><br />
```

### DIFF
--- a/src/Pgsql/PgsqlDriver.php
+++ b/src/Pgsql/PgsqlDriver.php
@@ -269,16 +269,19 @@ class PgsqlDriver extends PdoDriver
 		{
 			foreach ($fields as $field)
 			{
-				// Change Postgresql's NULL::* type with PHP's null one
-				if (preg_match('/^NULL::*/', $field->Default))
+				if ($field->Default !== null)
 				{
-					$field->Default = null;
-				}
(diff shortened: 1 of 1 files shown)


alikon:
we have 2 pr for this issue the 1st #258

alikon:
can we have merged at least one please

alikon (approved):

wilsonge:
Oops my bad. Was going through drone issues last night, missed the other one

PR #260, opened 2022-01-22
[2.0] [PHP 8.1] Fix deprecated null parameter for string functions in the escape method of the drivers

richard67:
Pull Request for Issue #

### Summary of Changes

Analogue to PR https://github.com/joomla/joomla-cms/pull/36787 for the 3.10-dev branch of the CMS.

### Testing Instructions

A code review could be sufficient. Otherwise see https://github.com/joomla/joomla-cms/pull/36787 .

### Additional comments

Reading the code I think the return value should be type casted to sting also for the `is_int($text)` case a bit above the `is_float($text)` modified here. (There is no PHP 8.1 error at that place but it would be right due to the function documentation).

But that would change behaviour regarding the return value's type in that case, and so I think this should be done with a future clean-up, together with type-hinting the methods.

None.

### DIFF
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -728,7 +728,7 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 		if (\is_float($text))
 		{
 			// Force the dot as a decimal point.
-			return str_replace(',', '.', $text);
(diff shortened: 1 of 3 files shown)


richard67:
Should I make an analogue PR for the 1.0-dev branch?

nibra:
Sorry for the thumb-down, was an accident :(

PR #262, opened 2022-02-19
Introduce Database aware trait and interface

laoneo:
Pull Request for Issue #26.

### Summary of Changes
Introduces a database aware interface and trait. Similar to what we have in other packages like the DI one.

### Testing Instructions
Run the unit tests.

New interface should be documented.

### DIFF
--- a/src/DatabaseAwareInterface.php
+++ b/src/DatabaseAwareInterface.php
@@ -16,6 +16,16 @@ namespace Joomla\Database;
 interface DatabaseAwareInterface
 {
 	/**
+	 * Get the database.
+	 *
+	 * @return  DatabaseInterface
+	 *
+	 * @since   2.0.3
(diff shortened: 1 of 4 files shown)


src/DatabaseAwareInterface.php
```diff
@@ -0,0 +1,28 @@
...
+	 * Set the database.
+	 *
+	 * @param   DatabaseInterface  $db  The database.
+	 *
+	 * @return  void
+	 *
+	 * @since   2.0.3
+	 */
+	public function setDatabase(DatabaseInterface $db): void;
+}
```
nibra, line 38: `protected getDatabase()` is missing

laoneo: The DI package has also not such a function. I would follow the same patter across the packages. In the CMS we are using getDbo() anyway for backwards compatibility.

laoneo: Done

PR #264, opened 2022-06-01
Database exporter fix for php 8.1 deprecations

alikon:
### Summary of Changes
used null coalescing operator

### Testing Instructions
with Error Reporting setted to Maximum
run `php cli/joomla.php database:export --folder=tmp/`

you got  a lot of `PHP Deprecated:  htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated in /shared/httpd/test/zzzz/libraries/vendor/joomla/database/src/DatabaseExporter.php on line 305`

### DIFF
--- a/src/DatabaseExporter.php
+++ b/src/DatabaseExporter.php
@@ -302,7 +302,7 @@ abstract class DatabaseExporter
 				{
 					if (!in_array($key, $colblob))
 					{
-						$buffer[] = '    <field name="' . $key . '">' . htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '</field>';
+						$buffer[] = '    <field name="' . $key . '">' . htmlspecialchars($value ?? '', ENT_COMPAT, 'UTF-8') . '</field>';
 					}
 					else
 					{


richard67:
Question is what will happen if we apply the fix and run an export and the import the exported file again? Will the import work and apply null values where necessary where we now will have empty strings with the fix?

nibra:
Whether you print null or an empty string does not make a difference. In both cases, the result will be `<field ...></field>`.

wilsonge:
We may need to differentiate though as those are different things in e.g. postgres

nibra:
> We may need to differentiate though as those are different things in e.g. postgres

Not in this case - it is about ` htmlspecialchars()` not liking null as a parameter.

Llewellynvdm:
@mbabker or @nibra what is the way forward here, I see more `null` issues and its probably only going to get worse. I already from @mbabker response on the registry issue can see we need to go over all of the framework to solve it. Almost seems like a [RECTOR](https://github.com/rectorphp/rector) task, to at least to get a dry run report of the scale of the task.

Anyway any more feedback here will help, thanks!

nibra:
In this case, the patch is ok and doing it's job.

mbabker:
No, it's not.  As George already pointed out, there is a huge difference between a database column storing a null value versus an empty string, and this patch makes it impossible to distinguish the two in an exported dataset.

nibra:
Although you're right that NULL *should* be handled differently, does this patch not change the behaviour. The exporter has always delivered an empty string for NULL values.

Hackwar:
I've created an alternative PR which properly handles NULL values: #286

alikon:
closing in favour of #264
better to have 1 pr

PR #268, opened 2022-08-17
Adding missing abstract getColumnSQL method to importer

Hackwar:
phpstan complains that we are accessing this method in the DatabaseImporter class, but it isn't defined, not even as abstract class. All importer classes have this method declared, so this is just a method of adding this after the fact. The question is, if this is acceptable from a semver point of view. An importer wouldn't be working without having this method implemented...


Hackwar:
We are going to fix this in 3.x. Probably.

PR #270, opened 2022-08-20
simplify PHP requirements

Hackwar:
I would assume that we support all PHP versions and with the nightly builds, we should learn early enough if something breaks in a future PHP version. That's why I'd like to simplify this the proposed way.


nibra:
The PHP requirement is that way on purpose - for now, only PHP <= 8.2 should be valid. 8.2 is tested with --ignore-platform-reqs, so it will work, once all dependencies are fixed. Your change exludes 8.1+

PR #271, opened 2022-09-20
[3.x] Other schema

softarius:
Pull Request for Issue #322 

### Summary of Changes
Support other then default schema for columns getting

### Testing Instructions
In database:
create schema foo;
create table foo.test (id integer, name varchar(10));

In Joomla:
Call getTableColumns method for this table 'foo.test'.

### DIFF
--- a/src/Pgsql/PgsqlDriver.php
+++ b/src/Pgsql/PgsqlDriver.php
@@ -220,6 +220,31 @@ class PgsqlDriver extends PdoDriver
     {
         return [];
     }
+    
+	/**
+	 * Parses the fully qualified or bare name of the database object (table, view, procedure etc.) into object 
+	 * with fields "schema" and "table" respectively
+	 * 
+	 * @param string $table Name of the database object with schema or without it
+	 * 
+	 * @return object
+	 */
+	private function ParseIdent($table)
+	{
+		$query=$this->getQuery(1);
+		$query->select(['r[1]','r[2]'])
+		->from("parse_ident(:t) as r")
+		->bind(':t', $table);
+		$qn=$this->setQuery($query)
+		->loadRow();
+		if ($qn[1])
+		{
+			return (object) ['schema' => $qn[0], 'table' => $qn[1]];
+		} else
+		{
+			return (object) ['schema' => $this->getDefaultSchema(), 'table' => $qn[0]];
+		}
+	}
 
     /**
      * Retrieves field information about a given table.
@@ -238,7 +263,7 @@ class PgsqlDriver extends PdoDriver
 
         $result        = [];
         $tableSub      = $this->replacePrefix($table);
-        $defaultSchema = $this->getDefaultSchema();
+        $fullname      = $this->ParseIdent($tableSub);
 
         $this->setQuery(
             '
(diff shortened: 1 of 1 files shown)


richard67:
As far as I can see by review, this PR will not work if either the schema name or the table name or both contain a dot, which is quite valid when the name which has a dot is name quoted. E.g. `"my.schema"."my.table"` is a valid schema, table combination.

Another question is why does this PR add that functionality only to the getTableColumns method and not to any other method which takes a table name as argument.

So in my opinion this PR should not be merged as it is. But other opinions are welcome.

richard67:
@softarius It seems you had rebased your branch locally to 3.x-dev and pushed the changes, but the PR on GitHub was still against the 2.0-dev branch. That's why it has shown so many unrelated changes.

I have fixed that now by rebasing to 3.x-dev on GitHub. Now it look ok again.

As far as I can see you have also resolved my previous comment. I will check and review again when I have more time.

src/Pgsql/PgsqlDriver.php
```diff
@@ -221,6 +221,29 @@ public function getTableCreate($tables)
...
+     */
+    private function parseIdent($table)
+    {
+        $query = $this->getQuery(1);
+        $query->select(['r[1]','r[2]'])
+        ->from("parse_ident(:t) as r")
+        ->bind(':t', $table);
+        $qn = $this->setQuery($query)
+        ->loadRow();
+        if ($qn[1]) {
```
richard67:
```suggestion
        $query = $this->getQuery(1);

        $query->select(['r[1]','r[2]'])
            ->from("parse_ident(:t) as r")
            ->bind(':t', $table);

        $qn = $this->setQuery($query)
            ->loadRow();

        if ($qn[1]) {
```
Better code style

src/Pgsql/PgsqlDriver.php
```diff
@@ -221,6 +221,29 @@ public function getTableCreate($tables)
...
+            return (object) ['schema' => $qn[0], 'table' => $qn[1]];
+        } else {
+            return (object) ['schema' => $this->getDefaultSchema(), 'table' => $qn[0]];
+        }
```
richard67:
```suggestion
        }

        return (object) ['schema' => $this->getDefaultSchema(), 'table' => $qn[0]];
```
No need for `else` if `if` code ends with return.

src/Pgsql/PgsqlDriver.php
```diff
@@ -363,6 +387,7 @@ public function getNamesKey($table, $indKey)
         $this->connect();
 
         $tableSub = $this->replacePrefix($table);
+        $fullname  = $this->parseIdent($tableSub);
```
richard67:
```suggestion
        $tableSub  = $this->replacePrefix($table);
        $fullname  = $this->parseIdent($tableSub);
```
PHPCS

src/Pgsql/PgsqlDriver.php
```diff
@@ -529,12 +555,14 @@ public function renameTable($oldTable, $newTable, $backup = null, $prefix = null
 
         $oldTable = $this->replacePrefix($oldTable);
         $newTable = $this->replacePrefix($newTable);
+        $fullOldTable = $this->parseIdent($oldTable);
+        $fullNewTable = $this->parseIdent($newTable);
```
richard67:
```suggestion
        $oldTable     = $this->replacePrefix($oldTable);
        $newTable     = $this->replacePrefix($newTable);
        $fullOldTable = $this->parseIdent($oldTable);
        $fullNewTable = $this->parseIdent($newTable);
```
PHPCS

src/Pgsql/PgsqlDriver.php
```diff
@@ -221,6 +221,32 @@ public function getTableCreate($tables)
...
+            ->bind(':t', $table);
+
+        $qn = $this->setQuery($query)
+            ->loadRow();
+
+        if ($qn[1]) {
+            return (object) ['schema' => $qn[0], 'table' => $qn[1]];
+        }
+
+        return (object) ['schema' => $this->getDefaultSchema(), 'table' => $qn[0]];
```
richard67, line 247: Here we get a 2nd database query in the `getDefaultSchema` method. That could cause some overhead.

The query here and the one in the `getDefaultSchema` method could be combined here so we won't need to call that method here and cause a 2nd query.

richard67, line 247:
```suggestion
        $query->select(['current_schemas(false))[1]','r[1]','r[2]'])
            ->from("parse_ident(:t) AS r")
            ->bind(':t', $table);

        $qn = $this->setQuery($query)
            ->loadRow();

        if ($qn[2]) {
            return (object) ['schema' => $qn[1], 'table' => $qn[2]];
        }

        return (object) ['schema' => $qn[0], 'table' => $qn[1]];
```
This should work to save us the additional 2nd query.

PR #273, opened 2023-01-13
php 8.2 - dynamic property deperecation

alikon:
Deprecated: Creation of dynamic property Joomla\Database\Mysql\MysqlDriver::$charset is deprecated in /shared/httpd/dev4/joomla-cms/libraries/vendor/joomla/database/src/Mysql/MysqlDriver.php on line 128

Pull Request for Issue #

### DIFF
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -94,6 +94,14 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	];
 
 	/**
+	 * The default charset.
+	 *
+	 * @var    string
+	 * @since  2.0.0
+	 */
+	protected $charset = 'utf8';
+
+	/**
 	 * Constructor.
 	 *
 	 * @param   array  $options  Array of database options with keys: host, user, password, database, select.
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -99,7 +99,7 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
 	 * @var    string
 	 * @since  2.0.0
 	 */
-	protected $charset = 'utf8';
+	public $charset = 'utf8';
 
 	/**
 	 * Constructor.


alikon:
@nibra can you check this one

HLeithner:
the variable has to be public since it's now public to, it can be deprecated and changed to protected in the next major version. beside that this variable is never used (at least I didn't found a usage). show it could be removed completely in next release?

nibra:
Thank you!

HLeithner:
and whats your plan for this useless variable?

richard67:
This fix is included in the 3.0.0 release, but not in the 2.1.1 because that was made in 2022, long before this PR here (and possibly others) were merged into 2.0-dev. So in Joomla 4 we get the deprecated warning when using the PDO MySQL driver on PHP 8.2, and this then results in errors when error reporting is set to maximum.

So it will need a new release 2.1.2 from the 2.0-dev branch first if we want to fix the problem with Joomla 4.4.1. @Hackwar do you think you can do that? It's not urgent, can be done after the 4.4.0 stable.

For removing the unused variable as suggested by @HLeithner above it's too late now, also for the 3.x-dev branch as we already have a 3.0.0 release (where this PR is included).

alikon:
i'm still bitten by this 

> It's not urgent, can be done after the 4.4.0 stable.

we have 4.4.3 now

richard67:
> i'm still bitten by this
> 
> > It's not urgent, can be done after the 4.4.0 stable.
> 
> we have 4.4.3 now

@alikon Not my fault.

@Hackwar Could you make a 2.1.2 release? Or do you know who can?

alikon:
@richard67 
it's not about fault, it's about how we can try to fix it

ThatsJustCheesy:
Will this change make it into a Joomla 4.x release? Without it, sites completely break on PHP 8.2+ when error reporting is "maximum".

> Deprecated: Creation of dynamic property Joomla\Database\Mysql\MysqlDriver::$charset is deprecated in ...

The deprecation warning preempts Joomla core setting HTTP headers, which makes all requests simply crash with "headers already sent by MysqlDriver.php"

HLeithner:
> Will this change make it into a Joomla 4.x release? 

No Joomla 4 is in security only mode.

PR #274, opened 2023-01-17
Replace getQuery(true) with createQuery(), deprecate $new parameter

HLeithner:
Since using a boolean parameter is a bad pattern and setting the parameter to true to get a new object is miss understandable and makes it hard for new comers to use the API.

The pull requests create a new method `createQuery()` which always returns a new databasequery. It also depreacte and mark the true parameter for removal in version 4.0

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -580,11 +580,11 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	/**
 	 * Create a new DatabaseQuery object.
 	 *
-	 * @return  DatabaseQuery
+	 * @return  QueryInterface
 	 *
-	 * @since   2.1
+	 * @since   2.2
 	 */
-	public function createQuery()
+	public function createQuery(): QueryInterface
 	{
 			return $this->factory->getQuery($this->name, $this);
 	}
@@ -1048,11 +1048,11 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	 * Get the current query object or a new DatabaseQuery object.
 	 *
 	 * @param   boolean  $new  False to return the current query object, True to return a new DatabaseQuery object.
+	 *                         The $new parameter is deprecated in 2.2 and will be removed in 4.0, use createQuery() instead.
 	 *
 	 * @return  DatabaseQuery
 	 *
 	 * @since   1.0
-	 * @deprecated  4.0  $new parameter will be removed, use createQuery() method instead.
 	 */
 	public function getQuery($new = false)
 	{
@@ -1060,8 +1060,9 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 		{
 			trigger_deprecation(
 				'joomla/database',
-				'2.1.0',
-				'getQuery(true) call is deprecated and will be removed in 4.0, use createQuery() instead.',
+				'2.2.0',
+				'The parameter $new is deprecated and will be removed in 4.0, use %s::createQuery() instead.',
+				self::class
 			);
 
 			return $this->createQuery();
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -578,6 +578,18 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	}
 
 	/**
+	 * Create a new DatabaseQuery object.
+	 *
+	 * @return  DatabaseQuery
+	 *
+	 * @since   2.1
+	 */
+	public function createQuery()
+	{
+			return $this->factory->getQuery($this->name, $this);
+	}
+
+	/**
 	 * Disconnects the database.
 	 *
 	 * @return  void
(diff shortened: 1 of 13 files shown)


laoneo:
I think the tests should not be changed. Instead add new test functions which do test the new function.

HLeithner:
The functionality it self is tested at https://github.com/joomla-framework/database/blob/2.0-dev/Tests/DatabaseFactoryTest.php#L261-L272

if we keep getQuery(true) in the test they will issue deprecation warnings. Maybe a new test for ->getQuery(true) which expects an deprecation warning make sense.

In v3 the interface should also get the new method.

wilsonge:
I don't think having a boolean param is inherently bad. But in this case given it's the majority use case it definitely makes sense to separate this so :+1:

src/DatabaseDriver.php
```diff
@@ -577,6 +577,18 @@ public function createDatabase($options, $utf = true)
 		return $this->execute();
 	}
 
+	/**
+	 * Create a new DatabaseQuery object.
+	 *
+	 * @return  DatabaseQuery
```
SharkyKZ: This should really be `QueryInterface`. The doc in `getQuery()` should be fixed too.

src/DatabaseDriver.php
```diff
@@ -577,6 +577,18 @@ public function createDatabase($options, $utf = true)
...
+	 *
+	 * @since   2.1
```
SharkyKZ: 2.1 is already out. Does this repo not support `__DEPLOY_VERSION__` placeholder?

src/DatabaseDriver.php
```diff
@@ -577,6 +577,18 @@ public function createDatabase($options, $utf = true)
...
+	 */
+	public function createQuery()
```
SharkyKZ: Add return type.

src/DatabaseDriver.php
```diff
@@ -1040,12 +1052,19 @@ public function getImporter()
 	 * @return  DatabaseQuery
 	 *
 	 * @since   1.0
+	 * @deprecated  4.0  $new parameter will be removed, use createQuery() method instead.
```
SharkyKZ: The method is not deprecated...

src/DatabaseDriver.php
```diff
@@ -1040,12 +1052,19 @@ public function getImporter()
...
 	 */
 	public function getQuery($new = false)
 	{
 		if ($new)
 		{
-			return $this->factory->getQuery($this->name, $this);
+			trigger_deprecation(
+				'joomla/database',
+				'2.1.0',
```
SharkyKZ: 2.1 is already out...

src/DatabaseDriver.php
```diff
@@ -1040,12 +1052,19 @@ public function getImporter()
...
+				'getQuery(true) call is deprecated and will be removed in 4.0, use createQuery() instead.',
```
SharkyKZ: Use full method names. Also, it's the parameter, not the method that's deprecated. Remove trailing comma.

src/DatabaseExporter.php
```diff
@@ -280,7 +280,7 @@ protected function buildXmlData()
 			}
 
 			$this->db->setQuery(
-				$this->db->getQuery(true)
+				$this->db->createQuery()
```
SharkyKZ: Revert.

src/Pgsql/PgsqlExporter.php
```diff
@@ -134,7 +134,7 @@ protected function buildXmlData()
 				}
 			}
 
-			$query = $this->db->getQuery(true);
+			$query = $this->db->createQuery();
```
SharkyKZ: Revert.

SharkyKZ (changes requested):

src/DatabaseExporter.php
HLeithner: Why?

src/Pgsql/PgsqlExporter.php
HLeithner: Why? It requests a new database query isn't it?

src/DatabaseDriver.php
HLeithner: Changed, not sure about the docs in `getQuery()`, it's completely wrong, because it could return the `interface` or a `string` (and base on `$this->sql` it could be `mixed`)

src/DatabaseDriver.php
HLeithner: Since jenkins is gone there is no bump script anymore afaik...

src/DatabaseDriver.php
HLeithner: done

src/DatabaseExporter.php
SharkyKZ: You can't use this here until the method is added to the interface.

HLeithner: true

HLeithner:
@nibra @Llewellynvdm would we like to merge this?

Llewellynvdm:
Please re-base, and I will merge this.

HLeithner:
on 3?, I think adding it to a 2.x minor makes the b/c better for extensions developer

wilsonge:
I agree adding this into the 2.x minor makes more sense to help with migration paths

wilsonge (changes requested):
I think we should leave one test in using `getQuery(true)` to test the legacy behavior still works (happy with moving the majority of test to the new style)

HLeithner:
Created a new test based on loadAssoc using the deprecated syntax.

Llewellynvdm:
Yes I was think 2 as well.

wilsonge (approved):

PR #276, opened 2023-01-23
Fix composer setup for Drone

Hackwar:
Composer is loading the wrong dependencies when --ignore-platform-reqs is used. This PR removes that option for PHP 8.0 and 8.1.


HLeithner:
thanks

PR #277, opened 2023-01-24
Cleanup of Appveyor configuration

Hackwar:
This PR cleans up the appveyor configuration and does the following:
- Update the image to Visual Studio 2022 (in sync with the CMS)
- Remove the caching, since it doesn't work. :-(
- Comment the build for PHP 8.1 and 8.2 because there are no pecl sqlsrv driver for these versions.
- Don't downgrade chocolatey anymore
- Properly install composer


Hackwar:
I was able to enable PHP 8.1 again.

PR #278, opened 2023-03-20
Updating dependencies, drone setup and adding phan setup

Hackwar:
### Summary of Changes
This PR removes phpmd from the drone setup and instead adds phan for additional static code analysis. It also updates the dependencies to the latest necessary versions and sets PHP 8.2 to be acceptable for drone.


Llewellynvdm (approved):
Okay since our build in drone is failing at this point, I will merge this to see ff it works.

Llewellynvdm:
What can we do to fix 2.0-dev?

PR #281, opened 2023-04-04
Update SqlsrvDriver.php

Kenneth202212:
### Summary of Changes

MS Changed their most recent SQLServer component for PHP to fail on SQL's default self-signed certificate unless you say TrustServerCertificate true. Unfortunately there is no current means to pass that value through the Joomla framework

### Testing Instructions

`

	$option = array();
	$option['driver']   = 'sqlsrv';
	$option['host']     = '(host)';
	$option['user']     = '(user)'; 
	$option['password'] = '(pass) ';
	$option['database'] = '(db-name)';
	$option['trust_certificate'] = true;

	$db = (new Joomla\Database\DatabaseFactory)->getDriver('sqlsrv', $option);

	$db->setQuery('SELECT COUNT(*) FROM MyTable');
	echo 'Results: ' . (int) $db->loadResult();`

### DIFF
--- a/src/Sqlsrv/SqlsrvDriver.php
+++ b/src/Sqlsrv/SqlsrvDriver.php
@@ -87,6 +87,7 @@ class SqlsrvDriver extends DatabaseDriver
 		$options['password'] = $options['password'] ?? '';
 		$options['database'] = $options['database'] ?? '';
 		$options['select']   = isset($options['select']) ? (bool) $options['select'] : true;
+		$options['trust_certificate'] = $options['trust_certificate'] ?? false;
 
 		// Finalize initialisation
 		parent::__construct($options);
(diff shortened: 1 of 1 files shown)


src/Sqlsrv/SqlsrvDriver.php
```diff
@@ -87,6 +87,7 @@ public function __construct(array $options)
 		$options['password'] = $options['password'] ?? '';
 		$options['database'] = $options['database'] ?? '';
 		$options['select']   = isset($options['select']) ? (bool) $options['select'] : true;
+		$options['trust_certificate'] = $options['trust_certificate'] ?? false;
```
richard67, line 90: This should be `$options['trust_certificate']   = isset($options['trust_certificate']) ? (bool) $options['trust_certificate'] : false;` so that it's casted to boolean if it is set as a string in the same way as it is done for the `$options['select']` above.

richard67: Another thing is that this change is not in the 3.x-dev branch. @rdeutz Should I make a PR to add this (in the corrected way as suggested by me above) in the 3.x-dev branch?

richard67: See #333 for my 1st and #332 for my 2nd comment above.

PR #283, opened 2023-06-08
Several fixes from phpstan

Hackwar:
### Summary of Changes
This PR has several changes from reports from phpstan.
1. The exception, which we are trying to catch, isn't thrown in the code. Instead an UnknownArchiveException is used and should thus be properly caught.
2. Several docblocks have been fixed.
3. `DatabaseImporter` got a signature for `getColumnSQL()`
4. `DatabaseInterface` got a signature for `getPrefix()`
5. `QueryInterface` got a signature for `quoteName()`

### DIFF
--- a/src/Command/ImportCommand.php
+++ b/src/Command/ImportCommand.php
@@ -12,6 +12,7 @@
 namespace Joomla\Database\Command;
 
 use Joomla\Archive\Archive;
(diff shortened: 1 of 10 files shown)


wilsonge:
The Interface modifications are b/c breaks which need documentation

PR #286, opened 2023-08-29
Exporter: Handle NULL values properly

Hackwar:
### Summary of Changes
Right now we are not handling NULL values properly in export and import. We are effectively converting them from NULL to an empty string, which is actually a big problem. This PR adds a new attribute to the exported data `value_is_null`. If that attribute is set, on import it is replaced with NULL.

Right now, this also creates lots of errors in PHP 8.1, because htmlspecialchars() should not be called with a NULL value.

### Testing Instructions
On PHP 8.1 in Joomla call the CLI database exporter and see the notices. Apply the changes and instead of the notices see the new attribute. Works especially well in Joomla 5.

This is a change in behavior, however for current backups the behavior would still be the same as before. Instead of a NULL value, you get an empty string. Also new exports would still be possible to be imported by old importers with the same behavior as before. The correct behavior to keep the NULL values would only work when both sides had this new feature and otherwise would simply behave like right now.

### DIFF
--- a/src/DatabaseExporter.php
+++ b/src/DatabaseExporter.php
@@ -302,9 +302,12 @@ abstract class DatabaseExporter
 				{
 					if (!in_array($key, $colblob))
 					{
-						if (is_null($value)) {
+						if (is_null($value))
+						{
 							$buffer[] = '    <field name="' . $key . '" value_is_null></field>';
-						} else {
+						}
+						else
+						{
 							$buffer[] = '    <field name="' . $key . '">' . htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '</field>';
 						}
 					}
--- a/src/DatabaseExporter.php
+++ b/src/DatabaseExporter.php
@@ -302,7 +302,11 @@ abstract class DatabaseExporter
 				{
 					if (!in_array($key, $colblob))
 					{
-						$buffer[] = '    <field name="' . $key . '">' . htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '</field>';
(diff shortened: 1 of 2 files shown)


src/DatabaseExporter.php
```diff
@@ -302,7 +302,11 @@ protected function buildXmlData()
 				{
 					if (!in_array($key, $colblob))
 					{
-						$buffer[] = '    <field name="' . $key . '">' . htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '</field>';
+						if (is_null($value)) {
+							$buffer[] = '    <field name="' . $key . '" value_is_null></field>';
```
richard67, line 307:
```suggestion
							$buffer[] = '    <field name="' . $key . '">value_is_null</field>';
```
From reading the code I would expect the `value_is_null` being the value without a leading space and not a part of the field tag just after name.

richard67: But I might be wrong.

richard67: Reading deeper, I think I'm wrong.

Hackwar: Yes, you are wrong. I definitely don't want to introduce some magic strings as values, which at some weird point will create yet another bug report. This is simply an empty attribute in the XML, nothing more.

richard67: Yes, I got that a few seconds after I had written my first comment. Unfortunately it seems I can't mark that as resolved here.

Hackwar: I already marked it as resolved.

Issue #290, opened 2023-11-12
With Mariadb 11.1.2 ordering of table is broken


tkuschel:
Copy from https://issues.joomla.org/tracker/joomla-cms/42333
https://github.com/joomla/joomla-cms/issues/42333
---
### FURTHER BUG REPORT

I took a closer look at the function `reorder` and found out that the query is created  with a windows function which is defined in `/libraries/vendor/joomla/database/src/DatabaseQuery.php` with the following line in

`public function selectRowNumber($orderBy, $orderColumnAlias)`

```
return $this->select("ROW_NUMBER() OVER (ORDER BY $orderBy) AS $orderColumnAlias");
```
but with 
`MysqlQueryBuilder.php` it is overwritten to
```
return $this->select("(SELECT @rownum := @rownum + 1 FROM (SELECT @rownum := 0) AS r) AS $orderColumnAlias");
```

This no longer works, so the first variant does it quite correctly.
Since the `MysqlQueryBuilder.php` is just a "traited" class of the former class `DatabaseQuery` in `DatabaseQuery.php`, we can simply remove this function in the `MysqlQueryBuilder.php`: 	(here it is commented out)

```

/*
  public function selectRowNumber($orderBy, $orderColumnAlias)
  {
     $this->validateRowNumber($orderBy, $orderColumnAlias);

     return $this->select("(SELECT @rownum := @rownum + 1 FROM (SELECT @rownum := 0) AS r) AS $orderColumnAlias");
  }
*/
```
It works for:

* MariaDB Version 11.1.2, MariaDB Version 10.11.4, 
* Both connections via MySQLi and PDO
* Joomla 5.0.0 and 4.4.0 with removed function `selectRowNumber` in `MysqlQueryBuilder.php`

Good to know:
The used SQL function `ROW_NUMBER()`, is introduced in MYSQL since version 8.0 (2016/09)
So the requirements for Joomla 5 with MySQL	8.1 minimum 8.0.13, and 
MariaDB	11.1.0 minimum 10.4.0 should work.

alikon:
please test #291

tkuschel:
> please test #291

who?

alikon:
to all whom would like this bug fixed
with some spare time & MariaDB 😄

richard67:
See also my alternative PR #300 which still supports older databases, as the minimum database requirements of the database framework have not been changed in the recent years so they are still the same on all 3 branches.

richard67:
[...]

PR #291, opened 2023-11-12
FIX  With Mariadb 11.1.2 ordering of table is broken #290

tkuschel:
Pull Request for Issue #290 
also see https://issues.joomla.org/tracker/joomla-cms/42333

### Summary of Changes
Removed   `public function selectRowNumber($orderBy, $orderColumnAlias)` in `MysqlQueryBuilder.php`
it already works within `DatabaseQuery.php` with window function `ROW_NUMBER()`
### Testing Instructions
see https://issues.joomla.org/tracker/joomla-cms/42333

only valid for Joomla 5.x because MariaDB and MySQL miniumum

### DIFF
--- a/src/Query/MysqlQueryBuilder.php
+++ b/src/Query/MysqlQueryBuilder.php
@@ -196,29 +196,6 @@ trait MysqlQueryBuilder
     }
 
     /**
-     * Return the number of the current row.
-     *
-     * Usage:
-     * $query->select('id');
-     * $query->selectRowNumber('ordering,publish_up DESC', 'new_ordering');
-     * $query->from('#__content');
-     *
-     * @param   string  $orderBy           An expression of ordering for window function.
-     * @param   string  $orderColumnAlias  An alias for new ordering column.
-     *
-     * @return  $this
-     *
-     * @since   2.0.0
-     * @throws  \RuntimeException
-     */
-    public function selectRowNumber($orderBy, $orderColumnAlias)
-    {
-        $this->validateRowNumber($orderBy, $orderColumnAlias);
-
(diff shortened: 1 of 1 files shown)


tkuschel:
Please be kind and add a test with the Mariadb 11.1.x version, see https://mariadb.com/kb/en/mariadb-11-1-3-release-notes/

PhilETaylor:
This issue DOES effect Joomla 3 sites running on `11.2.2-MariaDB` also. 

The fix to remove the function `selectRowNumber` in the Joomla 3 file `libraries/joomla/database/query/mysqli.php` just like this PR proposes for Joomla 5

nielsnuebel:
I have tested this successfully. We need this in the next Joomla Version asap.

tecpromotion:
I have tested this successfully. We need this in the next Joomla Version asap.

HLeithner:
We can't do this change because it would break the supported mysql versions for 3.x branch.

Can you please add a version check to only use the "parent" implementation when mariadb >= 11.0 is used?

tkuschel:
>Can you please add a version check to only use the "parent" implementation when mariadb >= 11.0 is used?

I'm currently on rehabilitation and can't work on it from here. Perhaps someone else can take over. I'll be away for at least 6 weeks.

richard67:
> I'm currently on rehabilitation and can't work on it from here. Perhaps someone else can take over. I'll be away for at least 6 weeks.

@tkuschel My best wishes for you. I hope all will go well. Maybe @HLeithner or me can take over.

richard67:
Regarding the versions check see my alternative PR #300 .

richard67:
~~See also PR #308 for another alternative way to do it.~~ Closed in favour of #300 .

rdeutz:
I have merged https://github.com/joomla-framework/database/pull/300. Thanks for providing a patch to fix the issue.

PR #294, opened 2023-12-21
[3.x-dev] New methods to get bool and string session variables

richard67:
Pull Request for new feature

Related to PR #293 .

Related CMS issues https://github.com/joomla/joomla-cms/issues/39479 and https://github.com/joomla/joomla-cms/issues/41156 .

### Summary of Changes

This Pull Request (PR) adds 2 new methods to get the values of session variables to the database drivers, one for boolean and one for string session variables.

This could be extended in future by methods to get array session variables like e.g. `sql_mode`.

For integer session variables this would not be easily applicable because the value range of integer session variables at least on MySQL and MariaDB exceeds everything which can be handled in PHP with integer or float variables without any maths extension. But they can be handled like strings.

This PR will allow Joomla CMS to show the values of the `sql_big_selects` and `max_join_size` variables in its system information.

The corresponding PR for the CMS is https://github.com/joomla/joomla-cms/pull/42559 .

### Testing Instructions

Will be added soon.

None.

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1893,4 +1893,34 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 
         return true;
     }
+
+    /**
+     * Get value of boolean session variable with the given name.
+     * Returns null if database doesn't support session variables or variable doesn't exist.
+     *
+     * @param   string  $name  The name of the session variable.
+     *
+     * @return  boolean|null
+     *
+     * @since   __DEPLOY_VERSION__
+     */
+    public function getSessionVarBool($name)
+    {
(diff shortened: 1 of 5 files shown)


richard67:
Meanwhile I was able to identify the critical SQL query in the CMS core which causes the issue with "max_join_size", and a pull request is ready for being merged: https://github.com/joomla/joomla-cms/pull/42576 . So for the CMS core the issue will be solved, and it will not need the changes from this PR here.

As it has turned out, the SQL error happens only on MariaDB but not on MySQL databases. It seems they handle the "max_join_size" differently.

That makes me think that the change proposed with this PR here is not really necessary, so I close it.

PR #297, opened 2024-02-27
getTableList() return only tables not views

alikon:
Pull Request for Issue #
follow up from https://github.com/joomla/joomla-cms/issues/42772

### Summary of Changes
exclude views from query results of `getTableList()`  same as postgres

### Testing Instructions
see https://github.com/joomla/joomla-cms/issues/42772

### DIFF
--- a/src/Mysql/MysqlDriver.php
+++ b/src/Mysql/MysqlDriver.php
@@ -487,7 +487,7 @@ class MysqlDriver extends PdoDriver implements UTF8MB4SupportInterface
         $this->connect();
 
         // Set the query to get the tables statement and not the views.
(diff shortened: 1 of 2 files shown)


src/Mysql/MysqlDriver.php
```diff
@@ -486,8 +486,8 @@ public function getTableList()
     {
         $this->connect();
 
-        // Set the query to get the tables statement.
-        return $this->setQuery('SHOW TABLES')->loadColumn();
+        // Set the query to get the tables statement and not the views.
```
richard67, line 489: Either revert the change on this comment, or do the same change on the same comment in the other file, so both comments are consistent.

alikon: done same comment

HLeithner:
Merged as decided by the maintainer meeting.

PR #300, opened 2024-03-16
[3.x] Use row_number window function on MariaDB >= 11.0

richard67:
Pull Request for Issue #290 .

Alternative to PR #291 .

See also https://github.com/joomla/joomla-cms/issues/42333 .

### Summary of Changes

Use the `ROW_NUMBER()` window function for MariaDB 11.0.0 and newer.

That function is supported on MySQL since version 8.0.0, too, and on MariaDB since version 10.2.0, but in order to play safe this PR only changes that for MariaDB 11.0.0 and newer, as for older MariaDB versions or MySQL the old code still seems to work. I've added a `@todo` comment to the `selectRowNumber` method and a comment to the version check for MariaDB 11.0.0 to make that clear.

I could not find any unit tests for the `selectRowNumber` method, so no changes on unit tests with this PR.

### Testing Instructions

See https://github.com/joomla/joomla-cms/issues/42333 .

None.

### DIFF
--- a/src/Query/MysqlQueryBuilder.php
+++ b/src/Query/MysqlQueryBuilder.php
@@ -213,11 +213,11 @@ trait MysqlQueryBuilder
      */
     public function selectRowNumber($orderBy, $orderColumnAlias)
     {
-        $this->validateRowNumber($orderBy, $orderColumnAlias);
-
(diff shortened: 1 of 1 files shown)


richard67:
See PR #308 for an alternative way to do it.

rdeutz:
I have tested this and it works as expected

PR #308, opened 2024-08-03
[3.x] Use row_number window function on MySQL and MariaDB if supported

richard67:
Pull Request for Issue #290 .

Alternative to PRs #291 and #300 .

See also https://github.com/joomla/joomla-cms/issues/42333 .

### Summary of Changes

Use the `ROW_NUMBER()` window function in the `selectRowNumber` method on MySQL >= 8.0.0 and MariaDB >= 10.2.0.

Different to PR #300 , this one here extends the database driver and database interface by the new method `hasRowNumberSupport` to tell if the database engine supports the `ROW_NUMBER()` window function, and the `MysqlQueryBuilder` uses that method. The new method might be used by 3rd party extensions of the CMS, too.

The MySQL drivers set a new property `rownum` when connecting to the database, and the new `hasRowNumberSupport` method just returns the value of that property, so the checks for MariaDB and the version are not done with every call to the `selectRowNumber` query method like it is done in PR #300 , so it should be a bit faster.

Also different to PR #300 , this one here uses the `ROW_NUMBER()` window function whenever supported, i.e. as said with MySQL >= 8.0.0 and MariaDB >= 10.2.0. The other PR #300 does that only for MariaDB >= 11.0.0 in order to play safe and because issues where only reported for MariaDB 11, but I think that's not really necessary. Using the `ROW_NUMBER()` is more reliable than the "hack" with the SQL variable (`@rownum`).

I could not find any unit tests for the `selectRowNumber` method, so no changes on unit tests with this PR.

### Testing Instructions

See https://github.com/joomla/joomla-cms/issues/42333 .

None.

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -134,6 +134,14 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
     protected $options;
 
     /**
+     * True if the database engine supports the ROW_NUMBER() window function.
+     *
+     * @var    boolean
+     * @since  __DEPLOY_VERSION__
+     */
+    protected $rownum = true;
+
+    /**
      * The current SQL statement to execute.
      *
      * @var    mixed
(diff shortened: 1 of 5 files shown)


richard67:
Closing in favour of PR #300 .

PR #311, opened 2024-09-08
[3.x] UPD deprecated getQuery(true)

tkuschel:
I am currently in the process of replacing the obsolete function getQuery(true) with createQuery() in Joomla CMS, see https://github.com/joomla/joomla-cms/pull/42344 (work in progress, at the meantime it is a closed PR, but open again after testing here).
To make my life easier, I have added the interface to the database with the createQuery() on the one hand, on the other hand you can use a little trick to build in a backtrace in the output via trigger_deprecated(), so that with the help (I use code OSS under Linux Arch) of the debugger (DEBUG CONSOLE) you can immediately jump to the calling file. See attached screenshot.
Screenshot from 2024-09-08 09-22-00

The former deprecated message was also very unclear and I have reworded it.
Perhaps other deprecated messages can also be modified accordingly?

### Summary of Changes
1. Rewording of debug message for deprecation of getQuery() with $new parameter.
2. Add the function createQuery() to the DatabaseInterface, also comment the deprecation there.
### Testing Instructions
Debug messages at the DEBUG CONSOLE or log file for Deprecated API in Joomla CMS.
I don't know if this is not already sufficiently documented.

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -988,10 +988,12 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
     public function getQuery($new = false)
     {
         if ($new) {
+            $key = array_search(__FUNCTION__, array_column(debug_backtrace(), 'function'));
+            $btfile = 'Found in: ' . debug_backtrace()[$key]['file'] . ':' . debug_backtrace()[$key]['line'];
             trigger_deprecation(
                 'joomla/database',
                 '2.2.0',
-                'The parameter $new is deprecated and will be removed in 4.0, use %s::createQuery() instead.',
+                __FUNCTION__ . " with a \$new (true) parameter is deprecated and will be removed in 4.0, use %s::createQuery() instead. $btfile",
                 self::class
             );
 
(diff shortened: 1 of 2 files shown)


HLeithner:
Changing the Interface can't be done in 3.x, also the call to $this->getQuery(true); can't be changed here because it would also be a b/c break.

So this has to be moved to 4.0

tkuschel:
Can we add the printing backtrace file only? - for 3.x  -- this is a great help when debugging

HLeithner:
Not sure if this is a great idea, mainly for performance reasons, getting the stacktrace for each query call in production doesn't sounds great for me.

tkuschel:
You are right performance decreases, when dealing with that, I can manually include that code during the replacements in my code.
I'll keep that PR as reminder, you may close.

HLeithner:
I keep this open and added it to the next framework maintainers meeting.

HLeithner:
Move Interface change to 4.0 and don't change the deprecation message using backtrace.

Hackwar:
Thank you for your work. With the work in #325, this change here isn't exactly necessary anymore. I'm sorry that it took this long. I'm closing this PR.

Issue #309, opened 2024-08-28
Method for determining database version fails if database is supplied through ProxySQL


MarkRS-UK:
### Steps to reproduce the issue
Try to install/update a J! site where the database is provided via ProxySQL and supplies some version(s) below the required limit. In the case of MySQL this is 5.5.30.

### Expected result
The installation/update is effected.

### Actual result
J! will report that it cannot make the required installation/update because the database version is too low.  It will report the database version as the lowest of the versions available via ProxySQL.

### Additional comments
This is discussed in some detail at https://joomlacommunity.cloud.mattermost.com/main/pl/hhouru9kopyx5f4amenpe3ksrr which also refers to the same issue discussed, and resolved, in the Drupal world at https://www.drupal.org/project/drupal/issues/3156078

richard67:
Am working on a fix.

richard67:
@MarkRS-UK Does the problem only happen when using “MySQL (PDO)” in The database settings of the Joomla installation? Or does it also happen when using “MySQLi”?

richard67:
The Drupal issue linked in the description of this issue here discussed possible solutions for getting the right version, but these were not implemented. The issue was closed with reference to its comment number 18, where another issue was linked which solves the issue by removing the code which had problems with the low version. This is something we can’t do.

alikon:
for my understaning this is a configuration issue and should be solved elsewhere

richard67:
> for my understaning this is a configuration issue and should be solved elsewhere

@alikon That's what was said first in the linked Drupal issue. But then it had turned out that the ProxySQL can support different versions and reports back the oldest one, and to fix that by configuration would mean that the ProxySQL would not allow to connect to these old versions anymore. Drupal fixed it by changing their code so it doesn't require a particular MySQL version anymore, which is something we cannot do. We could fix it by using e.g. `SELECT version();` like it was proposed in the Drupal issue at the beginning. The discussion in the Drupal issue doesn't really show me why that was rejected. I've tested that here and will provide a PR as soon as I know if it needs the fix only for MySQL (PDO) or also for MySQLi.

alikon:
the Joomla Technical Requirements are debatable to be honest, at least on db side, we require mysql 8 for J5 without any benefit, so my question is: how much ProxySQL is adopteed in the wild ? more than mysql 5.7 despite eol ?

richard67:
[...]

PR #312, opened 2024-09-08
[3.x] Get database server version from database so it works also with ProxySQL

richard67:
Pull Request for Issue #309

### Summary of Changes

This pull request (PR) changes the `getVersion` methods of the "MySQLi" and "MySQL (PDO)" drivers so that they select the version from the database after the first connection instead of relying on the connection's `server_info` in case of "MySQLi" or `PDO::ATTR_SERVER_VERSION` in case of "MySQL (PDO)".

The result is cached in a protected class variable so that the database query is only executed at the very first connection of the particular driver instance.

This makes it possible to use ProxySQL for a Joomla 5.x installation.

### Testing Instructions

With ProxySQL see issue #309 .

Otherwise (without ProxySQL) check that the version returned by `getVersion()` is the same as without this PR for the same database, e.g. something like "10.11.7-MariaDB-log" for MariaDB or "8.0.39-0ubuntu0.22.04.1" for MySQL databases.

None.

### DIFF
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -596,9 +596,9 @@ class MysqliDriver extends DatabaseDriver implements UTF8MB4SupportInterface
     }
 
     /**
-     * Get the version of the database connector.
+     * Get the version of the database server.
      *
-     * @return  string  The database connector version.
+     * @return  string  The database server version.
      *
      * @since   1.0
      */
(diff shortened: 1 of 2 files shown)


src/Mysqli/MysqliDriver.php
```diff
@@ -262,9 +270,11 @@ public function connect()
             $this->select($this->options['database']);
         }
 
-        $this->mariadb = stripos($this->connection->server_info, 'mariadb') !== false;
+        $serverVersion = $this->getVersion();
```
HLeithner, line 273: I'm not happy with this call on every connect, can we lazy load the mariadb and utf8mb4 variable?

richard67: That would be out of scope of my PR, I think. The change here is equal to what we already have in the MySQL PDO driver.

richard67:
Changing to draft due to review comments until I have a better solution.

PR #315, opened 2024-10-07
Allow IPv6 addresses with brackets for pgsql

muhme:
:construction: **DRAFT** as to be extended by unit tests and documentation
 
### Summary of Changes

IPv6 addresses are used with square brackets for MySQL and MariaDB and allowing the configuration of a non-default port number for the IPv6 address. However, with PostgreSQL, square brackets around IPv6 addresses are not usable, and setting a port is not possible.

Since it was decided to have only a single host field without a separate port number field in Joomla Web Installer (see [issue #43902](https://github.com/joomla/joomla-cms/issues/43902#issuecomment-2323932206)), there is a need to use square brackets to distinguish between the IPv6 host address and a port number.

This PR removes the square brackets somewhere in the middle, after separating a possible port number, but before continuing with PostgreSQL:
* PostgreSQL can now use IPv6 addresses with square brackets, just like MariaDB and MySQL → simplifies/unifies documentation and [joomla-cypress-36](https://github.com/joomla-projects/joomla-cypress/pull/36) implementation
* For the first time, it is possible to configure a port number together with an IPv6 address for PostgreSQL.
* The change is fully backward compatible. In the past, IPv6 addresses were accepted without square brackets, and this will continue to be supported in the future as well.

### Testing Instructions

:construction: **DRAFT** For the moment together with and will be added to [joomla-cypress-36](https://github.com/joomla-projects/joomla-cypress/pull/36) 

TODO

### DIFF
--- a/src/Pdo/PdoDriver.php
+++ b/src/Pdo/PdoDriver.php
@@ -261,6 +261,12 @@ abstract class PdoDriver extends DatabaseDriver
                 // Extract host and port or socket from host option
                 $this->setHostPortSocket(5432);
 
+                // PostgreSQL (in difference to MariaDB and MySQL) does not work with
+                // IPv6 addresses enclosed in square brackets, so we simply remove them here.
+                if (isset($this->options['host']) && substr($this->options['host'], 0, 1) === '[' && substr($this->options['host'], -1) === ']') {
+                    $this->options['host'] = trim($this->options['host'], '[]');
+                }
+
                 if ($this->options['socket'] !== null) {
                     $format = 'pgsql:host=#SOCKET#;dbname=#DBNAME#';
                 } else {


src/Pdo/PdoDriver.php
```diff
@@ -261,6 +261,12 @@ public function connect()
                 // Extract host and port or socket from host option
                 $this->setHostPortSocket(5432);
 
+                // PostgreSQL (in difference to MariaDB and MySQL) does not work with
+                // IPv6 addresses enclosed in square brackets, so we simply remove them here.
+                if (isset($this->options['host']) && substr($this->options['host'], 0, 1) === '[' && substr($this->options['host'], -1) === ']') {
+                    $this->options['host'] = trim($this->options['host'], '[]');
```
richard67, line 267: Unfortunately this modifies the host option in a way so that the colon in an IP v6 address with removed square brackets will cause a wrong port to be detected when the connect method runs a 2nd or 3rd or 4th ... time.

So it would need to use local variables here.

However, I have made an alternative PR #316 which does the bracket removal in the method to extract host, port and socket, controlled by a new parameter, and also handles the issue I mention here. My PR makes other refactoring in addition. Would be really appreciated if you could test my PR with your docker environments and system tests.

muhme: I haven’t encountered any issues running the overall System Tests. However, avoiding *hacks* in `options` looks better, and I’d close this PR once #316 is finished.

muhme:
This PR has been replaced by the alternative #316

PR #310, opened 2024-08-31
[3.x] Allow to specify port number or UNIX socket in host option also for MySQL (PDO) and PostgreSQL (PDO)

richard67:
Pull Request for CMS Issue https://github.com/joomla/joomla-cms/issues/43902 .

### Summary of Changes

This pull request (PR) makes it possible to specify a port number or UNIX socket in the host option for MySQL (PDO) and PostgreSQL (PDO) drivers in the same way as it is already possible for the MySQLi driver.

All these drivers already allow to specify a port number or a UNIX socket name with separate options.

But the CMS does not use these parameters. It does not have any fields in the configuration or installation form for that. Instead of that, it allows to append a port number to the host name separated by a colon, or to specify a UNIX socket with the `host` parameter.

The MySQLi driver contains code in its connect method to extract the port number or UNIX socket from the `host` parameter if that parameter contains one of these.

The MySQL (PDO) and PostgreSQL (PDO) drivers don't have such code.

It seems that for the MySQL (PDO) this is not a problem and it can handle such a `host` string as it is.

But the PostgreSQL (PDO) driver doesn't work with that, see the referred CMS issue.

This PR moves the code for extracting the port number or UNIX socket from the MySQLi driver's connect method to a new method in the base database driver so it can be used by any kind of database driver. The default port number is passed as a parameter to that method.

The PDO driver is changed so that it uses the new method for the MySQL (PDO) and PostgreSQL (PDO) drivers.

So it should work the same way for all database drivers used by the CMS: If the `host` option specifies a port number or UNIX socket, the port number or socket name from the `host` option are used (and override the corresponding parameters), otherwise port number or socket name from the options are used, if given, or the default port if nothing is given.

### Testing Instructions

See CMS issue https://github.com/joomla/joomla-cms/issues/43902 .

### Documentation Changes Required

None.

PR #316, opened 2024-10-27
[3.x] Remove square brackets from ipv6 host on PostgreSQL and refactor changes from PR 310

richard67:
Pull Request for Issue #

Alternative to #315 .

### Summary of Changes

This pull request (PR) adds an option to the `setHostPortSocket` method, which was added with PR #310 , to control if the square brackets around an IP v6 address in the `host` parameter should be kept ("MySQLi" and "MySQL (PDO)") or if they should be removed ("PostgreSQL (PDO)") in the connection parameters.

The default is true to keep the square brackets. For "PosgreSQL (PDO)" it is set to false so the quare brackets are removed.

This makes it possible to specify an IP v6 address together with a port number for "PostgreSQL (PDO)" in the same way as it already works for "MySQLi" and "MySQL (PDO)".

See also PR #315 for details.

In addition, this PR changes the `setHostPortSocket` method to not implicitly modify `$this->options` but to use function parameters and return value, and it renames that method to `extractHostPortSocket` to reflect that change.

The change to use function parameters and a return value makes it easier to provide unit tests for that method.

In the PDO driver, the original options are not modified but local variables are used. This makes sure that the options are not modified multiple times with each re-connect, which would cause problems with the removal of square brackets because the colons in an IP v6 address would cause wrong port number to be detected at the 2nd run.

In the MySQL driver this PR keeps it like it was so the behaviour is like before.

Finally, this PR adds a unit test for the `extractHostPortSocket` method.

### Request for comments - RFC

As the `extractHostPortSocket` method is protected, the unit test uses reflection to make it accessible.

An alternative would be to change that method to public.

@alikon @Hackwar @HLeithner @muhme and other reviewers: What do you think? Leave the unit test as it is? Or make the method public and test it without reflection?

### Testing Instructions

See CMS issue https://github.com/joomla/joomla-cms/issues/43902 and https://github.com/joomla-projects/joomla-cypress/pull/36 ,

And check in unit test logs in Drone if they contain a successful test "Pure host name or IP address and port or socket can be extracted from the host name option".

Don't know.

### DIFF
--- a/Tests/AbstractDatabaseDriverTestCase.php
+++ b/Tests/AbstractDatabaseDriverTestCase.php
@@ -1056,6 +1056,16 @@ abstract class AbstractDatabaseDriverTestCase extends DatabaseTestCase
         );
 
         $this->assertSame(
+            [null, null, '/path/to/unix/socket.sock'],
+            $refMethod->invoke(static::$connection, null, null, '/path/to/unix/socket.sock', 3306)
+        );
+
+        $this->assertSame(
+            [null, 3307, '/path/to/unix/socket.sock'],
+            $refMethod->invoke(static::$connection, null, 3307, '/path/to/unix/socket.sock', 3306)
+        );
+
+        $this->assertSame(
             [null, 3306, '/path/to/unix/socket.sock'],
             $refMethod->invoke(static::$connection, 'unix:/path/to/unix/socket.sock', null, null, 3306)
         );
--- a/Tests/AbstractDatabaseDriverTestCase.php
+++ b/Tests/AbstractDatabaseDriverTestCase.php
@@ -1036,4 +1036,193 @@ abstract class AbstractDatabaseDriverTestCase extends DatabaseTestCase
             $params
         );
     }
+
+    /**
+     * @testdox  Pure host name or IP address and port or socket can be extracted from the host name option
+     */
+    public function testExtractHostPortSocket()
+    {
+        $refObject = new \ReflectionObject(static::$connection);
+        $refMethod = $refObject->getMethod('extractHostPortSocket');
+
+        $this->assertSame(
+            ['', 3306, null],
+            $refMethod->invoke(static::$connection, '', null, null, 3306)
+        );
+
+        $this->assertSame(
+            ['', 3307, null],
+            $refMethod->invoke(static::$connection, '', 3307, null, 3306)
+        );
+
+        $this->assertSame(
+            [null, 3306, '/path/to/unix/socket.sock'],
+            $refMethod->invoke(static::$connection, 'unix:/path/to/unix/socket.sock', null, null, 3306)
+        );
+
+        $this->assertSame(
+            [null, 3307, '/path/to/unix/socket.sock'],
+            $refMethod->invoke(static::$connection, 'unix:/path/to/unix/socket.sock', 3307, null, 3306)
+        );
+
+        $this->assertSame(
+            ['192.168.67.254', 3306, null],
+            $refMethod->invoke(static::$connection, '192.168.67.254', null, null, 3306)
+        );
+
+        $this->assertSame(
+            ['192.168.67.254', 3307, null],
+            $refMethod->invoke(static::$connection, '192.168.67.254', 3307, null, 3306)
+        );
+
+        $this->assertSame(
+            ['192.168.67.254', 3308, null],
+            $refMethod->invoke(static::$connection, '192.168.67.254:3308', null, null, 3306)
+        );
+
+        $this->assertSame(
+            ['192.168.67.254', 3308, null],
+            $refMethod->invoke(static::$connection, '192.168.67.254:3308', 3307, null, 3306)
+        );
+
+        $this->assertSame(
+            ['[fe80:102::2%eth1]', 3306, null],
+            $refMethod->invoke(static::$connection, '[fe80:102::2%eth1]', null, null, 3306)
+        );
+
+        $this->assertSame(
+            ['[fe80:102::2%eth1]', 3307, null],
+            $refMethod->invoke(static::$connection, '[fe80:102::2%eth1]', 3307, null, 3306)
+        );
+
+        $this->assertSame(
+            ['[fe80:102::2%eth1]', 3308, null],
+            $refMethod->invoke(static::$connection, '[fe80:102::2%eth1]:3308', null, null, 3306)
+        );
+
+        $this->assertSame(
+            ['[fe80:102::2%eth1]', 3308, null],
+            $refMethod->invoke(static::$connection, '[fe80:102::2%eth1]:3308', 3307, null, 3306)
+        );
+
+        $this->assertSame(
+            ['fe80:102::2%eth1', 3306, null],
+            $refMethod->invoke(static::$connection, '[fe80:102::2%eth1]', null, null, 3306, false)
+        );
+
+        $this->assertSame(
+            ['fe80:102::2%eth1', 3307, null],
+            $refMethod->invoke(static::$connection, '[fe80:102::2%eth1]', 3307, null, 3306, false)
+        );
+
+        $this->assertSame(
+            ['fe80:102::2%eth1', 3308, null],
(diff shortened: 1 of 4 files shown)


richard67:
@alikon @muhme Could you test this one here? Just check if the changes from #310 are still working and if IP v6 addresses enclosed into square brackets are working. That would be much appreciated.

richard67:
@alikon @Hackwar @HLeithner @muhme and possible other other reviewers: Please check the "Request for comments - RFC" section in the description and let me know your opinion. Thanks in advance.

muhme:
Running System Tests with current `5.2-dev` and [JBT](https://github.com/muhme/joomla-branches-tester/) version 1.1.15 and bash command line:

1. Create 5.2-dev IPv6 installation with all needed patches
```
# joomla-cms-43968 and joomla-cms-44084 are already merged in 5.2
scripts/create 52 IPv6 database-310 database-316 joomla-cypress-33 joomla-cypress-36
```

2. Preparing tests with non-default database ports by setting up three port forwardings in the Cypress container:
```
docker exec jbt_cypress bash -c 'apt-get update && apt-get upgrade -y && apt-get install socat -y'
docker exec -d jbt_cypress socat 'TCP6-LISTEN:4711,fork,reuseaddr' 'TCP6:[fd00::11]:3306'
docker exec -d jbt_cypress socat 'TCP6-LISTEN:4712,fork,reuseaddr' 'TCP6:[fd00::12]:3306'
docker exec -d jbt_cypress socat 'TCP6-LISTEN:4713,fork,reuseaddr' 'TCP6:[fd00::13]:5432'
```

3. ✅ Test IPv6 with non-default port number using [test.sh](https://github.com/user-attachments/files/17559420/test.sh.txt) as *one-liner* script with Cypress configuration, the Web Installer installation step, and one test specification that uses custom database commands.

4. ✅  Test IPv6 with unset port number by adapting `test.sh` and running it again:

```
types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
hosts=("fd00::11" "fd00::11" "fd00::12" "fd00::12" "fd00::13")
ports=("" "" "" "" "")
```

5. ✅ Test IPv4 with unset port number by adapting `test.sh` and running it again:

```
types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
hosts=("10.0.0.11" "10.0.0.11" "10.0.0.12" "10.0.0.12" "10.0.0.13")
ports=("" "" "" "" "")
```

6. ✅ Test IPv4 with non-default port number by adapting `test.sh` and running it again:

```
types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
hosts=("10.0.0.7" "10.0.0.7" "10.0.0.7" "10.0.0.7" "10.0.0.7")
ports=("4711" "4711" "4712" "4712" "4713")
```

7. ✅ Test hostname with unset port number by adapting `test.sh` and running it again:

```
types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
hosts=("jbt_mysql" "jbt_mysql" "jbt_madb" "jbt_madb" "jbt_pg")
ports=("" "" "" "" "")
```

8. ✅ Test hostname with non-default port number by adapting `test.sh` and running it again:

```
types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
# Using docker container name and not 'jbt_cypress' as it’s an invalid hostname (will be fixed in the next JBT version).
hosts=("77b3b08bd2ff" "77b3b08bd2ff" "77b3b08bd2ff" "77b3b08bd2ff" "77b3b08bd2ff")
ports=("4711" "4711" "4712" "4712" "4713")
```

9. 🟥 Test Unix sockets by adapting `test.sh` and running it again:

```
types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
hosts=("unix:/var/run/mysql-socket/mysqld.sock" "unix:/var/run/mysql-socket/mysqld.sock" "unix:/var/run/mariadb-socket/mysqld.sock" "unix:/var/run/mariadb-socket/mysqld.sock" "unix:/var/run/postgresql-socket")
ports=("" "" "" "" "")
```

Failed for `mysqli` and `mariadbi` with 500 server error with:
```
Joomla\Database\DatabaseDriver::extractHostPortSocket(): Argument #1 ($host) must be of type string, null given, called in /var/www/html/libraries/vendor/joomla/database/src/Mysqli/MysqliDriver.php on line 203
```

richard67:
Hmm the MySQLi and the PDO driver set the host to „localhost“ in the constructor, so a null host value should never happen.

richard67:
Will check later why that fails.

muhme:
To verify, I created a new installation using the `database-315` patch instead, ran test No.9 Unix sockets and it is successfully passed.

richard67:
Yes, I know what the problem is. The method can run multiple times when connecting for checking the database and then disconnecting and connecting again to make the installation.

In the old code before PR #310 the `$this->options['host']` set to null at the first run and then was just passed again to the preg_replace at the 2nd run, which causes a deprecation notice on newer PHP version but not an error.

Now we have a function where the parameter is not nullable.

There are 2 ways to solve it: Either I do it like for PDO and do not modify the `$this->options` but use local variables, so we have the same starting conditions with every new connect. Or I make the `$host`parameter nullable and do an early return of the unmodified parameters when that parameter is null.

I tend to the first way.

The real clean way would be to use the method not in the connect method but in the constructor so it would run only one time on a particular instance.

richard67:
Hmm, on the other hand, the 2nd way would make it behave on MySQL like it was before PR #310 .

richard67:
>     9. 🟥 Test Unix sockets by adapting `test.sh` and running it again:
> 
> 
> ```
> types=("MySQLi" "MySQL (PDO)" "MySQLi" "MySQL (PDO)" "PostgreSQL (PDO)")
> hosts=("unix:/var/run/mysql-socket/mysqld.sock" "unix:/var/run/mysql-socket/mysqld.sock" "unix:/var/run/mariadb-socket/mysqld.sock" "unix:/var/run/mariadb-socket/mysqld.sock" "unix:/var/run/postgresql-socket")
> ports=("" "" "" "" "")
> ```
> 
> Failed for `mysqli` and `mariadbi` with 500 server error with:
> 
> ```
> Joomla\Database\DatabaseDriver::extractHostPortSocket(): Argument #1 ($host) must be of type string, null given, called in /var/www/html/libraries/vendor/joomla/database/src/Mysqli/MysqliDriver.php on line 203
> ```

@muhme Fixed. Thanks for testing so far.

muhme:
✅ Successfully completed the test with all 9 steps again, confirming that all tests passed smoothly this time. 👍

HLeithner:
I have some questions.

1. It seems hard to validated a ipv4 or ipv6 address with all it's implications, maybe using a lib for it is a better approach or at least look at it and make own code better, my first hit was https://github.com/mlocati/ip-lib no idea if it's good but it also support ipv6 zones
2. What's the reason todo the detection on connect and not in the constructor? If I saw it right we have no function to change the options later, and if we have we might simple run an HostPortWhateverupdate function with the new DSN string?

wouldn't this reduce a bit of complexity? at least I read somewhere something about double decode issue.

richard67:
@HLeithner Regarding your question 2: I don’t know why it is done in the connect method. It was like that since long time, if not since ever, when I checked the history on the 1.0-dev branch. I did not want to change that in a patch version. I fully agree that it should be done in the constructor. If there is consensus about that I can change this PR here (or make a new one).

Regarding question 1: Same reason as question 2, I did not want to make such a big change now.

We could discuss that in the next maintainers meeting for framework topics.

richard67:
I've created an alternative PR which makes the same changes as this one here (except of adding a unit test) plus moves the modification of options from the connect method to the constructor.

Closing in favour of #317 . Thanks for feedback and tests.

PR #324, opened 2025-03-05
[3.x] mysqli::ping is deprecated since php8.4

AkameOuO:
Pull Request for Issue #

### Summary of Changes
replace `ping()` with `stat() !== false`.
### Testing Instructions
MysqliDriver::connected() should work.

### DIFF
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -375,7 +375,7 @@ class MysqliDriver extends DatabaseDriver implements UTF8MB4SupportInterface
     public function connected()
     {
         if (\is_object($this->connection)) {
-            return $this->connection->ping();
+            return $this->connection->stat() !== false;
         }
 
         return false;


HLeithner:
Need to be checked if this is an performance issue, because this function is called for each sql query. So the overhead needs to be checked.

AkameOuO:
Tested on localhost with php 8.4.4 and mysql 8.1.
ping() spent 0.343s for 10000 calls and stat() spent 0.342s for 10000 calls.
I think it's ok

HLeithner:
Can you please test on the network with a second Computer please.

AkameOuO:
Ok. I Tested again with a remote database on AWS and local php server.
ping() spent 19.122s for 100 calls and stat() spent 19.110s for 100 calls.

P.S. It's about 190ms ping to the remote computer

HLeithner:
ok thanks, seems that ping did the same as stat();

HLeithner:
but much more interesting would be to get rid of this completely, Just thinking about using [mysqli_report](https://www.php.net/manual/en/function.mysqli-report.php) in exception mode and catch missing connections an retry after connect.... maybe another pr

brianteeman:
just came across this when running the joomla integration tests

HLeithner (approved):

rdeutz (approved):

richard67:
Thanks @AkameOuO and reviewers.

PR #326, opened 2025-03-16
feature: Update DatabaseFactory.php to allow external DB

alexandreelise:
Hey Joomla Team!
As I stumble on this problem, I would like to contribute with this tiny PR. TLDR; It allows to use the DatabaseFactory to inject external database for example in Joomla Tables e.g. ` $this->setDbo($externalDatabaseDriver);` 

*Before* this change it was not possible due to the fact that the return type of getDriver method required `DatabaseInterface` and the Parameter type of `setDbo` is exactly DatabaseDriver only. Maybe due to something related to Covariance and Contra-Variance in PHP.

*After* This change, since DatabaseDriver is an Abstract Class that implements DatabaseInterface it does not break the contract and normally everything should still work fine. But with the befenit for developers to use the full power of Joomla and the hard work of other Joomla Framework maintainers.

Pull Request for Issue #

### DIFF
--- a/src/DatabaseFactory.php
+++ b/src/DatabaseFactory.php
@@ -25,12 +25,12 @@ class DatabaseFactory
      * @param   string  $name     Name of the database driver you'd like to instantiate
      * @param   array   $options  Parameters to be passed to the database driver.
      *
-     * @return  DatabaseInterface
+     * @return  DatabaseInterface|DatabaseDriver
      *
      * @since   1.0
      * @throws  Exception\UnsupportedAdapterException if there is not a compatible database driver
      */
-    public function getDriver(string $name = 'mysqli', array $options = []): DatabaseInterface
+    public function getDriver(string $name = 'mysqli', array $options = []): DatabaseInterface|DatabaseDriver
     {
         // Sanitize the database connector options.
         $options['driver']   = preg_replace('/[^A-Z0-9_\.-]/i', '', $name);


HLeithner:
This seems to be more an error of the cms and not of the framework?

alexandreelise:
> This seems to be more an error of the cms and not of the framework?

Not really the CMS because now the CMS uses more and more the framework and in this example the CMS Table uses the framework DatabaseDriver 

see https://github.com/joomla/joomla-cms/blob/689c33f0c5d33756f775b6f032162e7de09ca5e5/libraries/src/Table/Table.php#L19

That's why I proposed this tiny change.

alexandreelise:
For reference. I found this.
https://docs.joomla.org/Connecting_to_an_external_database

HLeithner:
It is, because the cms table should never been set to `DatabaseDriver` instead it should have been set to the `DatabaseDriverInterface`. This can only be fixed with a new `setDatabase()` method in the table.

But your use case should work anyway if your own external class abstracts from the `DatabaseDriver`

alexandreelise:
> It is, because the cms table should never been set to `DatabaseDriver` instead it should have been set to the `DatabaseDriverInterface`. This can only be fixed with a new `setDatabase()` method in the table.
> 
> But your use case should work anyway if your own external class abstracts from the `DatabaseDriver`

If I understand you correctly the CMS Table (and not my code) should have DatabaseInterface (like in the Joomla Framework) and not DatabaseDriver like it is actually in CMS Table

So the CMS code should look like 
```php

public function __construct($table, $key, DatabaseInterface $db, ?DispatcherInterface $dispatcher = null)
    {
//CMS Table code here
}


```

Just to mention I already solved that problem since yesterday before posting this PR using a Decorator of DatabaseFactory but It would be easier if the core had this tiny change. Anyway thanks for your time Harald. I'll continue for now. Can't afford to wait anymore. Happy coding.

HLeithner:
do you have your own driver?

alexandreelise:
> do you have your own driver?
no. Did not have to go that far. I proposed that PR based on what worked for me and made that PR to basically avoid doing complicated workarounds rather than this single word in DatabaseFactory. At the moment, until this PR is eventually merged someday, I use a "Decorator" or whatever it's called as a Pattern around DatabaseFactory to allow me to change only the getDriver method to add the functionality you see in this PR. That's it. Don't have to code my own driver...yet.

HLeithner:
It is not clear where you have got an error, because the driver you get (if it's from the framework) is a DatabaseDriver object and would fit into the setDbo() function without any changes.

Can you should me the php error message?

alexandreelise:
> It is not clear where you have got an error, because the driver you get (if it's from the framework) is a DatabaseDriver object and would fit into the setDbo() function without any changes.
> 
> Can you should me the php error message?

As I said in the first place. I am not using the Driver directly otherwise the factory made be Joomla Framework team would be useless.  I am using the DatabaseFactory provided in the Framework https://github.com/joomla-framework/database/blob/d13a7504471b4730a8107e3566c6fe6d476706fc/src/DatabaseFactory.php#L33  but again, the problem is that the getDriver method of the factory hardcoded DatabaseInterface as return type and not DatabaseDriver as expected by setDbo hence the problem and hence this PR. Anyway, I already solved my problem so anyone can take this PR or not. Happy coding. Have a delightful day.

HLeithner:
PHP can't return an interface it will always return an object, that's the reason I ask for the php error message.
getDriver will return a DatabaseDriver object.

SharkyKZ (approved):

HLeithner:
Im closing this since we have a PR in the CMS https://github.com/joomla/joomla-cms/pull/45165

Issue #327, opened 2025-03-17
Improper escaping of identifiers


nikosdion:
This package does not properly quote identifiers (table and column names) which contain the identifier delimiter character in them. Here are examples of valid identifier (table or column) names in each DBMS which are improperly quoted by Joomla Database:
* 
* MySQL. ``foo`bar`` is escaped as `` `foo`bar` `` instead of ``` `foo``bar` ```.
* PostgreSQL, SQLite `foo"bar` is escaped as `"foo"bar"` instead of `"foo""bar"`.
* MS/Azure SQL Server `foo[bar]` is escaped as `[foo[bar]]` instead of `[foo[bar]]]`.

As to whether these are allowed identifier names, yes, they are. For example, according to [MySQL's documentation](https://dev.mysql.com/doc/refman/8.4/en/identifiers.html) backticks are acceptable characters in table and column names. Not that any sane person would use them, but they _are_ allowed.

The current implementation raises two concerns.

First, the misquoted identifiers cause a SQL error.

Second, and extending from the first, improper escaping allows an attacker to execute arbitrary SQL. For example, assuming a MySQL database and software which chooses table or column names from improperly filtered (e.g. `raw` instead of `cmd` Joomla input filtering) user-supplied values an attacker can manipulate the resulting SQL query to inject and execute arbitrary SQL commands.

The fix is straightforward. The [`DatabaseDriver::quoteNameStr‎`](https://github.com/joomla-framework/database/blob/d13a7504471b4730a8107e3566c6fe6d476706fc/src/DatabaseDriver.php#L1602-L1606) method needs to be changed like so:

```php
if (\strlen($q) === 1) {
	if (\strpos($part, $q) !== false)
	{
		$part = \str_replace($q, $q . $q, $part);
	}

	$parts[] = $q . $part . $q;
} else {
	if (\strpos($part, $q[1]) !== false)
	{
		$part = \str_replace($q[1], $q[1] . $q[1], $part);
	}

	$parts[] = $q[0] . $part . $q[1];
}
```

Here's how this works in the supported database servers:

**MySQL**

Escape a backtick by writing it as two consecutive backticks. [Reference](https://dev.mysql.com/doc/refman/8.4/en/identifiers.html).

> Identifier quote characters can be included within an identifier if you quote the identifier. If the character to be included within the identifier is the same as that used to quote the identifier itself, then you need to double the character. The following statement creates a table named a`b that contains a column named c"d:
> ```mysql
[...]

PR #328, opened 2025-03-18
[2.x] Replace incorrect working quoteNameStr method

HLeithner:
Pull Request for Issue #327

### DIFF
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1702,27 +1702,7 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	 */
 	protected function quoteNameStr($strArr)
 	{
-		$parts = [];
-		$q     = $this->nameQuote;
-
-		foreach ($strArr as $part)
-		{
-			if ($part === null)
-			{
-				continue;
-			}
-
-			if (\strlen($q) === 1)
-			{
-				$parts[] = $q . $part . $q;
-			}
-			else
-			{
-				$parts[] = $q[0] . $part . $q[1];
-			}
-		}
-
-		return implode('.', $parts);
+		return $this->quoteNameString(implode('.', $strArr));
 	}
 
 	/**
--- a/src/DatabaseDriver.php
+++ b/src/DatabaseDriver.php
@@ -1702,7 +1702,19 @@ abstract class DatabaseDriver implements DatabaseInterface, DispatcherAwareInter
 	 */
 	protected function quoteNameStr($strArr)
 	{
-		return $this->quoteNameString(implode('.', $strArr));
+		$parts = [];
+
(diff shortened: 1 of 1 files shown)


SniperSister (approved):

ramalama (approved):

richard67:
The change of the `quoteNameStr` seems not to be fully b/c as there is a change in behaviour when the string to be quoted contains a dot and another difference when an array element is null.

When you call the old, unmodified, deprecated method with the following array as argument
```
['foo.bar','foo`bar','foo"bar','foo[bar','foo]bar',null]
```
you get with MySQL names quotes:
```
`foo.bar`.`foo`bar`.`foo"bar`.`foo[bar`.`foo]bar`
```
With this PR you get:
```
`foo`.`bar`.`foo``bar`.`foo"bar`.`foo[bar`.`foo]bar`.``
```
The same applies with other names quotes (PostgreSQL, MS SQL Server).

The first array element with the dot is quoted differently.

The second array element shows the fix for the issue.

The last array element shows the difference when null.

If that b/c break is a problem, I suggest following different implementation:
```
    function quoteNameStr($strArr)
    {
        $parts = [];

        foreach ($strArr as $part) {
            if ($part === null) {
                continue;
            }

            $parts[] = quoteNameString($part, true);
        }

        return implode('.', $parts);
    }
```
Mind the 2nd parameter of the `quoteNameString` call.

My suggested code gives following result for the example array:
```
`foo.bar`.`foo``bar`.`foo"bar`.`foo[bar`.`foo]bar`
```

The same applies when using PostgreSQL or MS SQL Server names quotes: With both ways the issue would be fixed, but with this PR the first array element with the dot is quoted differently.

You can test that here: https://onlinephp.io/c/26575

src/DatabaseDriver.php
```diff
@@ -1702,27 +1702,7 @@ protected function quoteNameString($name, $asSinglePart = false)
...
-				$parts[] = $q . $part . $q;
-			}
-			else
-			{
-				$parts[] = $q[0] . $part . $q[1];
-			}
-		}
-
-		return implode('.', $parts);
+		return $this->quoteNameString(implode('.', $strArr));
```
richard67:
```suggestion
		$parts = [];

		foreach ($strArr as $part)
		{
			if ($part === null)
			{
				continue;
			}

			$parts[] = quoteNameString($part, true);
		}

		return implode('.', $parts);
```
This would be fully b/c to the old deprecated method, also for dots in the string or for null.

HLeithner:
thanks richard

richard67:
@SniperSister @ramalama Could you check this PR again? Thanks in advance.

richard67 (approved):

SniperSister (approved):

PR #334, opened 2025-03-25
Don't include .git-blame-ignore-revs file in distribution package

AkameOuO:
### Summary of Changes

Same as #218 .

### Testing Instructions

The zip package will not include .git-blame-ignore-revs file.

None.


HLeithner:
Thx

PR #338, opened 2025-04-06
[3.x] Use $this->quoteNameString instead of quoteNameString

Hackwar:
### Summary of Changes
With the fix of the `quoteNameStr()` method, the `$this->` was forgotten. This PR fixes that.


richard67 (approved):
Was my silly mistake because I had suggested that code in my review.

Issue #254, opened 2021-09-01
$db->loadObjectList('id', MenuItem::class) fails on Unknown fetch type '8'


mahagr:
### Steps to reproduce the issue

This example needs Joomla, see issue https://github.com/joomla/joomla-cms/issues/35452

```php
use Joomla\CMS\Menu\MenuItem;

$db = \JFactory::getDbo();
$query = $db->getQuery(true)
    ->select('m.id, m.menutype, m.title, m.alias, m.note, m.path AS route, m.link, m.type, m.level, m.language')
    ->select($db->quoteName('m.browserNav') . ', m.access, m.params, m.home, m.img, m.template_style_id, m.component_id, m.parent_id')
    ->select('e.element as component')
    ->from('#__menu AS m')
    ->join('LEFT', '#__extensions AS e ON m.component_id = e.extension_id')
    ->where('m.menutype = ' . $db->quote($menutype))
    ->where('m.parent_id > 0')
    ->where('m.client_id = 0')
    ->where('m.published >= 0')
    ->order('m.lft');

// Set the query
$db->setQuery($query);

$list = $db->loadObjectList('id', MenuItem::class);
```

### Expected result

The query should work just as it worked in Joomla 3.9/3.10.

### Actual result

Exception: `Unknown fetch type '8'`. This seems to happen when using `mysqli` but not when using `dbo`.

Type 8 is same as `FetchMode::CUSTOM_OBJECT `.

### System information (as much as possible)

PHP 8.0.10 / 7.4.23

### Additional comments

Following workaround works (replace the last line):

```php
$list = [];
foreach ($db->loadAssocList('id') as $id => $data) {
    $list[$id] = new MenuItem($data);
}
```

mahagr:
The error happens here:

https://github.com/joomla-framework/database/blob/2.0.0/src/Mysqli/MysqliStatement.php#L542-L546

And this is how it is handled in a couple of other drivers:

https://github.com/joomla-framework/database/blob/2.0.0/src/Pdo/PdoStatement.php#L177
https://github.com/joomla-framework/database/blob/2.0.0/src/Sqlsrv/SqlsrvStatement.php#L464-L467

So it looks like `FetchMode::CUSTOM_OBJECT` is being handled everywhere else but in `mysqli` driver. Please check the other drivers, though, as I didn't take a look in every one of them.

mbabker:
[...]

PR #348, opened 2025-10-06
[4.0] Deprecated class attribute

rdeutz:
Pull Request for Issue #254 

### Summary of Changes
In Meeting 106 of the maintainers team the team decided to remove the class parameter from loadObject, loadObjectList and DatabaseIterator. 

This way (ussing a custom class as result) of fetching data is not supported over all database drivers so it can lead to problems in the development and using the database package.

### Testing Instructions
No tests possible it is ony documenting that the parameter is deprecated and will be removed with version 6 of the package.

No changes needed it is so widly used that it wasn't part of the documentation ;-)

### DIFF
--- a/src/DatabaseIterator.php
+++ b/src/DatabaseIterator.php
@@ -70,9 +70,11 @@ class DatabaseIterator implements \Countable, \Iterator
      * @param   StatementInterface  $statement  The statement holding the result set to iterate.
      * @param   string              $column     An option column to use as the iterator key.
      * @param   string              $class      The class of object that is returned.
+     *                                          The $class parameter is deprecated in 4.x and will be removed in 6.0
+     *                                          without a replacement.
      *
-     * @since   1.0
      * @throws  \InvalidArgumentException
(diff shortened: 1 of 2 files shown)


src/DatabaseIterator.php
```diff
@@ -70,9 +70,11 @@ class DatabaseIterator implements \Countable, \Iterator
      * @param   StatementInterface  $statement  The statement holding the result set to iterate.
      * @param   string              $column     An option column to use as the iterator key.
      * @param   string              $class      The class of object that is returned.
+     *                                          The $class parameter is deprecated in 4.x and will be removed in 6.0
+     *                                          without a replacement.
      *
-     * @since   1.0
      * @throws  \InvalidArgumentException
+     *@since   1.0
```
richard67:
```suggestion
     * @since   1.0
```

HLeithner:
I think this should trigger an custom deprecation message if the input is not \stdClass, since no IDE or analyser  understands parameter deprecation.

PR #349, opened 2025-10-17
Add mysqli exception mode

kamil-tekiela:
This PR changes the `MysqliDriver` to use exception mode just like PDO does. I tried not to make too many changes and stick to the previous style. There was some dead code in `executeUnpreparedQuery` which I removed to make the change easier.

### DIFF
--- a/src/Mysqli/MysqliDriver.php
+++ b/src/Mysqli/MysqliDriver.php
@@ -202,7 +202,10 @@ class MysqliDriver extends DatabaseDriver implements UTF8MB4SupportInterface
             throw new UnsupportedAdapterException('The MySQLi extension is not available');
         }
 
-        $this->connection = mysqli_init();
+        // Enable mysqli error reporting
+        mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
+
+        $this->connection = new \mysqli();
 
         $connectionFlags = 0;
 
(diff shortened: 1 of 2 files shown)


HLeithner:
thanks, maybe we can remove the warning suppression (@) now too? other question is can we get the state before we set it and set it back after connecting? 

It also seems that `MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT` is the default since php 8.1, that would mean we already get exceptions in this driver but doesn't catch them?

kamil-tekiela:
> thanks, maybe we can remove the warning suppression (@) now too? other question is can we get the state before we set it and set it back after connecting?

Thanks for pointing this out. I removed the error suppression. 
We can get the state from `mysqli_driver->report_mode` but I don't see why we would need to do that. 

> It also seems that `MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT` is the default since php 8.1, that would mean we already get exceptions in this driver but doesn't catch them?

Yes, that would be true. Hence, why the current code needs to be changed because it doesn't work as intended. I still like to set the error mode explicitly though.

rdeutz (approved):
code review

richard67 (approved):
Code review ok, and I've also done some tests on 5.4-dev of the CMS.

richard67:
Thanks @kamil-tekiela for this fix, and @HLeithner and @rdeutz for reviews.

PR #350, opened 2025-10-17
Remove all redundant references

kamil-tekiela:
This PR removes `call_user_func_array` which was necessary in PHP 5. The references magic was only needed for `call_user_func_array`. The only reference needed is for `rowBindedValues` which is because PHP cannot handle references to properties the same way as it does with variables. In the future version, you can get rid of this completely after switching to `get_result`.

### DIFF
--- a/src/Mysqli/MysqliStatement.php
+++ b/src/Mysqli/MysqliStatement.php
@@ -31,7 +31,7 @@ class MysqliStatement implements StatementInterface
      * @var    array
      * @since  2.0.0
      */
-    protected $bindedValues;
+    protected $bindedValues = [];
 
     /**
      * Mapping between named parameters and position in query.
@@ -39,7 +39,7 @@ class MysqliStatement implements StatementInterface
      * @var    array
      * @since  2.0.0
      */
-    protected $parameterKeyMapping;
+    protected $parameterKeyMapping = [];
 
     /**
      * Mapping array for parameter types.
@@ -301,27 +301,23 @@ class MysqliStatement implements StatementInterface
      */
     private function bindValues(array $values)
     {
-        $params = [];
         $types  = str_repeat('s', \count($values));
 
-        if (!empty($this->parameterKeyMapping)) {
-            foreach ($values as $key => &$value) {
+        if ($this->parameterKeyMapping !== []) {
+            $params = [];
+            foreach ($values as $key => $value) {
                 $paramKey = $this->parameterKeyMapping[$key];
                 foreach ($paramKey as $currentKey) {
-                    $params[$currentKey] =& $value;
+                    $params[$currentKey] = $value;
                 }
             }
 
             ksort($params);
-        } else {
-            foreach ($values as $key => &$value) {
-                $params[] =& $value;
-            }
+
+            return $this->statement->bind_param($types, ...$params);
         }
 
-        array_unshift($params, $types);
-
-        return \call_user_func_array([$this->statement, 'bind_param'], $params);
+        return $this->statement->bind_param($types, ...$values);
     }
 
     /**
@@ -372,22 +368,22 @@ class MysqliStatement implements StatementInterface
      */
     public function execute(?array $parameters = null)
     {
-        if ($this->bindedValues !== null) {
+        if ($this->bindedValues !== []) {
             $params = [];
             $types  = [];
 
-            if (!empty($this->parameterKeyMapping)) {
-                foreach ($this->bindedValues as $key => &$value) {
+            if ($this->parameterKeyMapping !== []) {
+                foreach ($this->bindedValues as $key => $value) {
                     $paramKey = $this->parameterKeyMapping[$key];
 
                     foreach ($paramKey as $currentKey) {
-                        $params[$currentKey] =& $value;
+                        $params[$currentKey] = $value;
                         $types[$currentKey]  = $this->typesKeyMapping[$key];
                     }
                 }
             } else {
-                foreach ($this->bindedValues as $key => &$value) {
-                    $params[]    =& $value;
+                foreach ($this->bindedValues as $key => $value) {
+                    $params[]    = $value;
                     $types[$key] = $this->typesKeyMapping[$key];
                 }
             }
@@ -395,9 +391,7 @@ class MysqliStatement implements StatementInterface
             ksort($params);
             ksort($types);
 
-            array_unshift($params, implode('', $types));
-
-            if (!\call_user_func_array([$this->statement, 'bind_param'], $params)) {
+            if (!$this->statement->bind_param(implode('', $types), ...$params)) {
                 throw new PrepareStatementFailureException($this->statement->error, $this->statement->errno);
             }
         } elseif ($parameters !== null) {
@@ -418,15 +412,7 @@ class MysqliStatement implements StatementInterface
             $meta = $this->statement->result_metadata();
 
             if ($meta !== false) {
-                $columnNames = [];
-
-                foreach ($meta->fetch_fields() as $col) {
-                    $columnNames[] = $col->name;
-                }
-
-                $meta->free();
-
-                $this->columnNames = $columnNames;
+                $this->columnNames = array_column($meta->fetch_fields(), 'name');
             } else {
                 $this->columnNames = false;
             }
@@ -436,13 +422,10 @@ class MysqliStatement implements StatementInterface
             $this->statement->store_result();
 
             $this->rowBindedValues = array_fill(0, \count($this->columnNames), null);
-            $refs                  = [];
+            // The following is necessary as PHP cannot handle references to properties properly
+            $refs =& $this->rowBindedValues;
 
-            foreach ($this->rowBindedValues as $key => &$value) {
-                $refs[$key] =& $value;
-            }
-
-            if (!\call_user_func_array([$this->statement, 'bind_result'], $refs)) {
+            if (!$this->statement->bind_result(...$refs)) {
                 throw new \RuntimeException($this->statement->error, $this->statement->errno);
             }
         }


src/Mysqli/MysqliStatement.php
```diff
@@ -301,24 +301,20 @@ public function bindParam($parameter, &$variable, string $dataType = ParameterTy
...
-        $params = [];
         $types  = str_repeat('s', \count($values));
 
-        if (!empty($this->parameterKeyMapping)) {
-            foreach ($values as $key => &$value) {
-                $params[$this->parameterKeyMapping[$key]] =& $value;
+        if ($this->parameterKeyMapping !== []) {
+            $params = [];
+            foreach ($values as $key => $value) {
+                $params[$this->parameterKeyMapping[$key]] = $value;
```
HLeithner: This will remove the reference between `$values` and `bind_param`

src/Mysqli/MysqliStatement.php
```diff
@@ -369,32 +365,30 @@ public function errorInfo()
...
-                        $params[$currentKey] =& $value;
+                        $params[$currentKey] = $value;
                         $types[$currentKey]  = $this->typesKeyMapping[$key];
                     }
                 }
             } else {
-                foreach ($this->bindedValues as $key => &$value) {
-                    $params[]    =& $value;
+                foreach ($this->bindedValues as $key => $value) {
+                    $params[]    = $value;
```
HLeithner, line 386: Same here, you create a copy of `$value` which should actually be a reference to the `$this->bindedValues`

src/Mysqli/MysqliStatement.php
```diff
@@ -415,15 +409,7 @@ public function execute(?array $parameters = null)
             $meta = $this->statement->result_metadata();
 
             if ($meta !== false) {
-                $columnNames = [];
-
-                foreach ($meta->fetch_fields() as $col) {
-                    $columnNames[] = $col->name;
-                }
-
-                $meta->free();
```
HLeithner, line 429: What's the reason you removed the `free()` call?

HLeithner (changes requested):
First thanks for your pull request. I added some comments.

I didn't tested if the reference with mysql statement object but tested it with user defined variable-length argument lists parameter which worked fine.

Can you elaborate what you mean with 

> after switching to `get_result`

src/Mysqli/MysqliStatement.php
kamil-tekiela: I am not entirely sure what you mean. Yes, it will remove the reference but that is the point. Are you saying that it was used for something else? The method is private and it's only called from within `execute` so the only way I see it could be used is if someone executed the statement with parameters and then relied on that binding instead of passing the parameters again when executing it again. But that surely could not have been intended behaviour (if it even works as I haven't tested it this way). If anything, I would consider it a bug fix. But maybe I am misunderstanding your comment.

src/Mysqli/MysqliStatement.php
kamil-tekiela: I am pretty sure this doesn't change anything, but I can't test it now because this feature is currently broken in your project. But again, maybe I am misunderstanding something. I will try to find the bug and patch up my local version to see if I can test it.

src/Mysqli/MysqliStatement.php
kamil-tekiela: `mysqli_result::free` should only be used when you need to close the object prematurely and the object came from a different scope (and even then, it's pretty dubious). The `$meta` variable is a local variable that never leaves the scope. Calling this method has no effect here (not that it ever would with `result_metadata` as there is no result to close).

kamil-tekiela:
> Can you elaborate what you mean with
>
> > after switching to get_result

Right now, composer says that PHP 8.1 is the minimum required version. From PHP 8.2 the `get_result` becomes always available. Until then, you cannot make the switch as it could potentially break code for people who are running mysqli compiled with libmysqlclient. 

Once you make the switch to PHP 8.2, you can replace the current code because you do no use result binding in reality. You just need it because there is no alternative.  Using `get_result` will simplify the code and remove the workaround left by me.

src/Mysqli/MysqliStatement.php
HLeithner: That's exactly how bind works for all database drivers. you execute and change the variable and execute again. 

Don't ask me why someone would use it that way (ok I used it already in some projects) but that's the reason statement->bind_param parameters are references. So also the reason why in the joomla framework the `bind` method are references (even if I hate it because I can't use function results at this point).

src/Mysqli/MysqliStatement.php
HLeithner: see other answer

src/Mysqli/MysqliStatement.php
HLeithner: Ok, I see and yes I know it get freed automatically if we get out of the scope, but if you search for this you might find some resources on the internet where if you don't close it you might get trouble with transaction. But didn't tested it my self to tbh.

HLeithner:
> > Can you elaborate what you mean with
> > > after switching to get_result
> 
> Right now, composer says that PHP 8.1 is the minimum required version. From PHP 8.2 the `get_result` becomes always available. Until then, you cannot make the switch as it could potentially break code for people who are running mysqli compiled with libmysqlclient.
> 
> Once you make the switch to PHP 8.2, you can replace the current code because you do no use result binding in reality. You just need it because there is no alternative. Using `get_result` will simplify the code and remove the workaround left by me.

I think I still don't get you sorry, which function is new in php 8.2? get_result on the statement exists since 5.3 and no changes are in the function description. Also we use bind_result if I have checked the code correctly.

src/Mysqli/MysqliStatement.php
kamil-tekiela: Sorry, I think I have caused some confusion with my statement. I just tried what I said above and it is impossible IMHO. I was referring to a pretty esoteric way of using binding. The binding you have in Joomla still works. This PR does not change the behaviour of how this class works (at least it shouldn't, so let me know if you have observed some changes).

As I said in the PR description, the `&` was only necessary for the `call_user_func_array`. They were not necessary for the binding functionality that you have implemented in Joomla. It was just a workaround because of that level of indirection added by `call_user_func_array`.

src/Mysqli/MysqliStatement.php
kamil-tekiela: I have fixed the named parameter binding in the other PR and I was able to test that no change occurs with this PR.

src/Mysqli/MysqliStatement.php
kamil-tekiela: I am not sure what transaction you are referring to, but I assure you: this line did nothing. Trust me. `result_metadata` doesn't produce an object that can be closed. But even with normal result objects, using `free()` is 99.99% redundant.

kamil-tekiela:
> > > Can you elaborate what you mean with
> > > > after switching to get_result
> > 
> > 
> > Right now, composer says that PHP 8.1 is the minimum required version. From PHP 8.2 the `get_result` becomes always available. Until then, you cannot make the switch as it could potentially break code for people who are running mysqli compiled with libmysqlclient.
> > Once you make the switch to PHP 8.2, you can replace the current code because you do no use result binding in reality. You just need it because there is no alternative. Using `get_result` will simplify the code and remove the workaround left by me.
> 
> I think I still don't get you sorry, which function is new in php 8.2? get_result on the statement exists since 5.3 and no changes are in the function description. Also we use bind_result if I have checked the code correctly.

There is no new function. As you can see on https://www.php.net/manual/en/mysqli-stmt.get-result.php in the note box:

> Available only with [mysqlnd](https://www.php.net/manual/en/book.mysqlnd.php).

And if you recall my RFC that was implemented in PHP 8.2 https://wiki.php.net/rfc/mysqli_support_for_libmysql the feature to compile mysqli with libmysqlclient was dropped. So since PHP 8.2, you can safely use `get_result()` knowing that it will be available on all user platforms. 

My whole point is that you don't need `bind_result` and your code only uses it because `get_result` might not have been available. In the future, you can refactor the code and remove `bind_result`.

src/Mysqli/MysqliStatement.php
HLeithner: I try to show you a valid example which breaks with this code changes:

```php
// We assume that we are in the Joomla context
$db = \Joomla\CMS\Factory::getContainer()->get('db');
$query = $db->createQuery();
$title = 'Title with ';
$query->insert('#__tablename')
            ->columns($db->quoteName(['title']))
            ->values(':title')
            ->bind(':title', $title);

        $db->setQuery($query);
        for($i=0; $i<10; $i++) {
            $title .= 'x';
            $db->execute();
        }
```

This will add 10 rows with the titles:
* Title with x
* Title with xx
* Title with xxx
* Title with xxxx
* Title with xxxxx
* Title with xxxxxx
* ....

Which is perfectly fine, that's how the mysql api is designed to use and also the joomla database framework works.

kamil-tekiela: What do you mean it breaks?

With just the bug fix PR #351:
```
Title with x
Title with xx
Title with xxx
Title with xxxx
Title with xxxxx
Title with xxxxxx
Title with xxxxxxx
Title with xxxxxxxx
Title with xxxxxxxxx
Title with xxxxxxxxxx
```
With both PRs:
```
Title with x
Title with xx
Title with xxx
Title with xxxx
Title with xxxxx
Title with xxxxxx
Title with xxxxxxx
Title with xxxxxxxx
Title with xxxxxxxxx
Title with xxxxxxxxxx
```

I'm going to try to be a little clearer. I worked with mysqli for years. I am familiar with PHP internals and I have fixed a number of bugs in mysqli. I know very well how mysqli works and I am familiar with the concept of binding parameters (both in PDO and mysqli). I have never used Joomla, but it seems straight-forward to figure out with my vast PHP experience. I know that Joomla is very popular and I know that the changes I and others have made in recent years to PHP are impacting Joomla. I have decided to open a PR to fix the issues I am aware of. 

I realize that Joomla exposes the concept of parameter binding to its users. Without Joomla experience, I humbly admit that I could have missed some little quirk when preparing the PR. But from my testing and the unit test I have added in the other PR, I can see that nothing has changed in the way Joomla works. All that is fixed is the old code is gone. 

Sorry for being so blunt, but I feel a little belittled. It seems to me like you are trying to teach me the concept of parameter binding while you haven't tested my PR. I apologize if it's otherwise. I **kindly and humbly** ask you to carefully review my PR again, with testing.

HLeithner: I didn't want to upset you, sorry for this and I only tested it theoretically, we don't have a complete reference from user land to `stmt->bind_param` function with your pr. But I tried your code now and didn't head in mind that the framework code calls `bind_param` again on each `execute`.

So you are completely right it's not needed to reference all the way down and it's enough to do only the reference to the `->bindedValues` property.

it would only be relevant if we don't do this... sorry for wasting your time.

src/Mysqli/MysqliStatement.php
HLeithner: it would directly call mysql_free_result right? so we can control when it gets freed, on long running processes that doesn't sound like a bad idea to free it early and don't wait for the GC.

HLeithner:
> And if you recall my RFC that was implemented in PHP 8.2 https://wiki.php.net/rfc/mysqli_support_for_libmysql the feature to compile mysqli with libmysqlclient was dropped. So since PHP 8.2, you can safely use `get_result()` knowing that it will be available on all user platforms.
> 
> My whole point is that you don't need `bind_result` and your code only uses it because `get_result` might not have been available. In the future, you can refactor the code and remove `bind_result`.

ok, now I get it, thanks. would be useful for the next major release. If you like I to create a pr I can create a new branch for it.

src/Mysqli/MysqliStatement.php
kamil-tekiela: Yes, it does call it but with `result_metadata` you only have a `mysqli_result` with metadata without any data. So it only frees metadata, which is stored in PHP memory anyway, ergo it does nothing. With a normal result, it would fetch all remaining data from an unbuffered result set, and it would discard stored data from a buffered result set. 

In either case, the code doesn't wait for GC. GC is only called when there are cyclic references. When there are not as in this case, the memory is freed up immediately when no longer pointed to by a variable. The `$meta` variable doesn't leave the scope, doesn't hold any data, and there is nothing major happening until the scope exits. 

There are some other places in the code where this is happening, and I will create another PR for it.

HLeithner: ok, make sense.

src/Mysqli/MysqliStatement.php
```diff
@@ -301,27 +301,22 @@ public function bindParam($parameter, &$variable, string $dataType = ParameterTy
      */
     private function bindValues(array $values)
     {
-        $params = [];
         $types  = str_repeat('s', \count($values));
 
-        if (!empty($this->parameterKeyMapping)) {
+        if ($this->parameterKeyMapping !== []) {
             foreach ($values as $key => &$value) {
```
richard67:
```suggestion
            foreach ($values as $key => $value) {
```
Should this be changed, too? It has landed here with my conflict resolution due to the merged PR #351 .

richard67: Seems I made another mistake when resolving the conflict as the `$params` variable is not defined with the changes from this PR here?

src/Mysqli/MysqliStatement.php
```diff
@@ -301,27 +301,22 @@ public function bindParam($parameter, &$variable, string $dataType = ParameterTy
      */
     private function bindValues(array $values)
     {
-        $params = [];
```
richard67, line 304: Should this be reverted?

src/Mysqli/MysqliStatement.php
kamil-tekiela: I have rebased. It should be ok now.

richard67: Thanks.

richard67:
@HLeithner GitHub still shows that you requested changes, but the review comments are all resolved. Could you review again and approve if ok? Thanks in advance.

HLeithner (approved):

rdeutz (approved):
code review

richard67:
@kamil-tekiela I've allowed myself to fix the 2 conflicts in the `src/Mysqli/MysqliStatement.php` file caused by the previously merged PR #349 . Please check and report back if I've done that right. Thanks in advance.

kamil-tekiela:
Sorry for the delay. Yes, that is correct.

HLeithner:
Thanks @kamil-tekiela

PR #351, opened 2025-10-18
Fix MysqliStatement to work with named params

kamil-tekiela:
It seems that some time ago the named parameter feature was adjusted to cover for multiple keys. This place was missed and has been broken ever since. 

@HLeithner This also covers the changes in #350. Both tests should still work after the references are removed.

### DIFF
--- a/Tests/Mysqli/MysqliPreparedStatementTest.php
+++ b/Tests/Mysqli/MysqliPreparedStatementTest.php
@@ -51,6 +51,14 @@ class MysqliPreparedStatementTest extends DatabaseTestCase
                 )
             );
         }
+
+        $insertQuery = 'INSERT INTO dbtest (title, description, start_date) VALUES (:title, :description, :start_date)';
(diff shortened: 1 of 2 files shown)


HLeithner:
Thanks, I think nobody noticed this because no one is using the statement class directly and the database driver doesn't accept any variables on `execute` method.

HLeithner (approved):

richard67 (approved):
Code review successful. It seems that place was forgotten with PR #303 . In the execute method it has made that change in the same way.

richard67:
I've successfully tested this PR by creating the table and running the code added by this PR here to the unit test.

Without this PR you get an error "Cannot access offset of type array on array", with line 309 of the `MysqliStatement.php` file at the top of the call stack, and the record is not inserted into the test table.

With this PR the record is inserted without any error.

richard67:
Thanks.

PR #353, opened 2025-10-18
Remove escape when not with string

kamil-tekiela:
Methods `quoteName` and `escape` cannot be used together as they serve different purposes. `escape` is for formatting strings in SQL, whereas `quoteName` is for formatting identifiers. Using both of them at the same time will lead to incorrect results.


HLeithner:
thanks

PR #357, opened 2026-02-15
Add bindValue method to DatabaseQuery

HLeithner:
You can't bind a literal or function result directly to the bind value method because it expects a variable.

### Summary of Changes
Add `bindValue` method which is a proxy to the `bind method.

### Testing Instructions
Only a function has been added, test the `bindValue` if it works as you expect `bind` to work without the reference part.

Not sure if we have a proper database documentation already, if yes it needs to be added

### Future
Add this method to the `DatabaseQueryInterface`

### DIFF
--- a/src/DatabaseQuery.php
+++ b/src/DatabaseQuery.php
@@ -1940,6 +1940,30 @@ abstract class DatabaseQuery implements QueryInterface
     }
 
     /**
+     * Method to add a variable to an internal array that will be bound to a prepared SQL statement before query execution.
+     * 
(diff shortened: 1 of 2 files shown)


src/DatabaseQuery.php
```diff
@@ -1939,6 +1939,30 @@ public function bind($key, &$value, $dataType = ParameterType::STRING, $length =
         return $this;
     }
 
+    /**
+     * Method to add a variable to an internal array that will be bound to a prepared SQL statement before query execution.
+     *
```
richard67:
```suggestion
     *
```
Fix PHPCS white space at end of line.

